PR #3, opened 2020-02-21
Adding examples for MAAP search functionality

KMarkert:
Added in a `search/granules.ipynb` example that walks users through using the maap-py `seachGranule` method and some parameters to support search.

Added in a page with a table that lists all of the collections within the MAAP CMR instance.

Removed the visualizing raw data example. This will be moved under a visualization directory at a later stage.


kvirts:
- [ ] L7: Current workbook titles are "searching for collections" and "searching granules."  I'd say "searching for granules."

- [ ] L9: The UWG members already know this, but for future users it might help to define what collections and granules are, and how they relate to each other.

- [ ] L9: "through"

- [ ] L18: "invoke the API"

- [ ] L30: "Here we view the specific arguments and keywords for the..."

- [ ] L62: Suggest ending first sentence after "results from CMR."  Then, "maap.searchGranule also accepts any additional search parameters that are included in CMR.  For a list of accepted parameters, please refer to..."

- [ ] L81: "we will explore search options".  Also note that this does not appear as a numbered list in the finished notebook.

- [ ] L86: "additional"

- [ ] L88: "additional"

- [ ] L95: Consider using consistent capitalization for major sub-headings.

- [ ] L97: "collection IDs" (no apostrophe)

- [ ] L97: The link to the collection table doesn't work (maybe because the page isn't "live" yet?)

- [ ] L99: "Collection"

- [ ] L99: In other contexts, users may want the ambiguity of not specifying a collection.  As in, what granules from other datasets also include information on this location that I'm studying?  Maybe a separate example for this at some point.

- [ ] L126: Perhaps "The output from the MAAP API is..."

- [ ] L128: "the first result"

- [ ] L140: Maybe have an example (here or elsewhere) of how to display the fields that are listed as {...} in the granule metadata.  (E.T.A. I guess they can infer this from some of the later commands.)

- [ ] L163: Is depth=2 something that should be explained?

- [ ] L172: "further"

- [ ] L183: Suggest "Temporal search criteria may be either a single date or a date range."

- [ ] L185: Either "one month of data" or "one month's worth of data"

- [ ] L202: For end date, should it be 2018-12-31T23:59:59Z?

- [ ] L225: Perhaps "was acquired starting at"

- [ ] L232: "the results are limited to 100"

- [ ] L241: Perhaps "height data over Southeast Asia"

- [ ] L258: Perhaps use consistent variable names, like maxResults and boundingBox (or max_results and bounding_box), etc.  I'm getting picky... :)

- [ ] L283: For the bounding box example, change variable name from granuleDate to granuleBoundingBox or granuleDomain.

- [ ] L302: "additional"

- [ ] L302: Maybe break up the second sentence with a period after "There are many additional attributes available."

- [ ] L304: These are good examples of additional attributes they might well use.

- [ ] L309: Is there a list of site names, or a way to tell how exactly they should be entered?  I can imagine a user searching for "Mondah Forest, Gabon" or just "Monday Forest"...

- [ ] L370: "additional" (twice)

- [ ] L377: "functionality"

KMarkert:
2. Added a description of a granule in the intro, please check to make sure it makes sense.

7. Please clarify what you would like changed to "we will explore search options". The whole line? Just a specific portion? 

12. Yes, the link is not live yet. This will work once pushed to the website.

14. Yes, this is another example to be completed later.

15. I am hesitant to replace "result" with "output" these are two different things and technically the result from the server is what we are referring to at this point

17. I added a comment for users to change the depth of the output. I left it at 2 so that there is not a lot of nonsense on the example.

18. depth is a keyword for controlling how deep within a dict tree the outputs are printed. I don't want to go into detail explaining everything if it is not functionality for maap. Users can easily look up what this means.

26. You are not getting picky. You are right that the variable naming should be consistent. In Python "constant" variables are all caps so I changed `MAX_RESULTS` to `MAXRESULTS`. All other variables should be [camelCase](https://en.wikipedia.org/wiki/Camel_case) please let me know if they are not.

31. I have no clue...this is still really opaque at least for the the site names. All other additional attributes are in the table we use to populate the metadata.

32. Where? I see "additional attributes" twice but I put that in there to be explicit.


I committed the changes please review my comments and fixes to make sure they are correct.

docs/source/search/granules.ipynb
```diff
@@ -0,0 +1,404 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Searching for Granules in MAAP\n",
+    "\n",
+    "These examples will walk through the MAAP API functionality of searching granules within a collection based on specific parameters. Granules can be thought of individual scenes from a sensor where a group of granules make a collection within CMR. The granules are the raw data that will be used for processing."
```
kvirts: Granule description looks good to me.

docs/source/search/granules.ipynb
```diff
@@ -0,0 +1,404 @@
...
+   "source": [
+    "# get at max 500 results from CMR\n",
+    "MAXRESULTS = 500"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this example we will explore search options including:\n",
```
kvirts: This looks good now.

docs/source/search/granules.ipynb
```diff
@@ -0,0 +1,404 @@
...
+    "\n",
+    "results = maap.searchGranule(concept_id=COLLECTIONID,limit=MAXRESULTS)\n",
+    "pprint(f'Got {len(results)} results')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We were able to get 500 results! There are most likely more than 500 granules in search results, but remember we limited the results to 500 granules. The results from the MAAP API is a list of granules where each element in the list is the metadata for that particular granule.\n",
```
kvirts: Ok to use result rather than output.  But it should be "The result...is a list"

docs/source/search/granules.ipynb
```diff
@@ -0,0 +1,404 @@
...
+      " 'collection-concept-id': 'C1200116818-NASA_MAAP',\n",
+      " 'concept-id': 'G1200125381-NASA_MAAP',\n",
+      " 'format': 'application/echo10+xml',\n",
+      " 'revision-id': '1'}\n"
+     ]
+    }
+   ],
+   "source": [
+    "# print the first granule metadata\n",
+    "# adjust the depth to a larger value (4) if you would like to view all of the metadata\n",
```
kvirts: This is good.

docs/source/search/granules.ipynb
```diff
@@ -0,0 +1,404 @@
...
+    "pprint(results[0],depth=2)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The returned results will give you only GeoTIFF datasets that have been tagged as part of the Mondah Forest Gabon research site.\n",
+    "\n",
+    "**DISCLAIMER:** The MAAP data team is working to update the addtional attributes within the MAAP platform so these are subject to change. Furthermore, the accepted parameters for the addtional attributes are changing and further documentation will be provided as updates come."
```
kvirts: "Additional" is misspelled twice in this paragraph.

kvirts:
These changes look good. I left clarification in two places; once they are fixed, I'm ready to merge.

docs/source/search/granules.ipynb
KMarkert: 🎉

docs/source/search/granules.ipynb
KMarkert: 🎉

docs/source/search/granules.ipynb
KMarkert: 🎉

docs/source/search/granules.ipynb
KMarkert: Just pushed updated. Please verify and resolve

docs/source/search/granules.ipynb
KMarkert: Just pushed updated. Please verify and resolve

docs/source/search/granules.ipynb
```diff
@@ -0,0 +1,404 @@
...
+    "granuleDate = results[0]['Granule']['Temporal']['RangeDateTime']['BeginningDateTime'] # get the granule start time\n",
+    "\n",
+    "pprint(f'Granule {granuleFilename} was acquired starting at {granuleDate}',width=100)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It looks like the first result correctly matches with the beginning temporal search parameter. Keep in mind that the results are limited to 100 so the final granule returned may not match the end date that was searched for."
```
kvirts: Just noticed that the 100 in this paragraph should be 500.

kvirts:
Sorry, I did a final read-through and noticed one more typo...

docs/source/search/granules.ipynb
KMarkert: Just committed the change

PR #20, opened 2020-07-01
Added GEDI Cal/Val Data document and updated the index

spa0002:
-  Added an ipython notebook example that explains the fields within the cal/val data including variable name, description and units.

   -  Within index.rst, added the GEDI Cal/Val data documentation notebook to the contents tree

### DIFF
--- a/docs/source/query/gedi_calval_data_doc.ipynb
+++ b/docs/source/query/gedi_calval_data_doc.ipynb
@@ -18,6 +18,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


docs/source/query/gedi_calval_data_doc.ipynb
```diff
@@ -0,0 +1,129 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Fields Within the Gedi Cal/Val Data"
```
spa0002, line 7: add an introductory paragraph above the table

PR #34, opened 2020-10-29
[WIP] Feat/wcs edav

wildintellect:
Example retrieving data from EDAV WCS service. Not sure if it's ready for the live site, maybe we need a staging branch.
cc: @abarciauskas-bgse


abarciauskas-bgse:
Did we mean to merge this already? It's great but wasn't sure it was quite ready for the docs website.

@wildintellect what do you think?

wildintellect:
@spa0002 and I discussed that it should be ok to have it in the Develop branch of the site, which is not the official definitive document (that's latest which comes from master branch). But it's an Architecture question for the group, where to put the in progress documentation.

PR #35, opened 2020-10-30
Updated query_data_python.ipynb to include example for `where` query

spa0002:
- query_data_python.ipynb has been updated to include example query with multiple parameters using `where`.

### DIFF
--- a/docs/source/query/query_data_python.ipynb
+++ b/docs/source/query/query_data_python.ipynb
@@ -183,30 +183,30 @@
       "{\n",
       "  \"project\": \"gabon_mondah\",\n",
       "  \"plot\": \"NASA11\",\n",
-      "  \"subplot\": \"1\",\n",
+      "  \"subplot\": \"8\",\n",
       "  \"survey\": \"AfriSAR_ESA_2016\",\n",
       "  \"tree.date\": \"2016-02-10\",\n",
-      "  \"family\": \"Burseraceae\",\n",
-      "  \"species\": \"Aucoumea klaineana\",\n",
+      "  \"family\": \"Euphorbiaceae\",\n",
+      "  \"species\": \"Maprounea membranacea\",\n",
       "  \"pft\": null,\n",
-      "  \"wsg\": 0.377851851851852,\n",
-      "  \"wsg.sd\": 0.0708232617216773,\n",
-      "  \"tree\": \"5503\",\n",
+      "  \"wsg\": 0.588,\n",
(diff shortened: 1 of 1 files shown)


abarciauskas-bgse:
@spa0002 this is awesome! I love the explanation and the examples. I would suggest we tweak the structure if you agree. I think we should restructure the documentation that's already a part of what is in this PR and at the top of the notebook to "introduce" the `executeQuery` function and detail it's parameters. Some useful information about poll_results and timeout is buried in the example you added. Below I detail what I'm imagning which merges with the existing content of `Query Searching for a Project Name` with your content:

## How to use `maap.executeQuery()`

We use the `executeQuery()` function to return a response object, containing the server's response to our HTTP request. This object can be used to view the response headers, access the raw data of the response, or parse the response as a JavaScript Object Notation (JSON). JSON is a data-interchange format, designed to be easy for humans to read and write.

### `executeQuery` parameters

* `src` - dictionary-like object specifying the dataset to query. At the moment the only valid option is:

```
{ 
  "Collection": { 
      "ShortName": "GEDI Cal/Val Field Data_1", 
      "VersionId": "1" 
  } 
}
```
* `query` dictionary-like object specifying parameters for query. Dictionary can include `bbox`, `where` and `fields`:
    * `bbox` - list of floats identifying a bounding box of geographic coordinates
    * `where` - dictionary of fields and corresponding values. Used to query the data set for rows with matching values for the corresponding fields.
    * `fields` - a list of fields to return, a subset of all fields available for the corresponding dataset.
* `poll_results`:  Parameter which must be `True` to use the timeout parameter
* `timeout`: Waiting period for a response. Indicates the maximum number of seconds to wait for a response. Note that `timeout` has a default value of '180' and requires that the `poll_results` parameter be `True`. Depending on the request, it may be necessary to modify the timeout to make sure the server has enough time to process the request.

### Query Searching for a Project Name
In this example, we assign our `collection` dictionary to `src`, a dictionary-like object specifying the dataset to query. We also assign the `bbox` and `fields` parameters for `query`, ...


Let me know if the above makes sense or I can add a commit to show what I am thinking.

abarciauskas-bgse:
FYI I'm working on some changes here https://github.com/MAAP-Project/maap-documentation/tree/abarciauskas-bgse_v1d2-demo for both this notebook and the ATL08 notebook. I can make a PR to this branch if it helps @spa0002

spa0002:
Hi Aimee,

Thanks for the suggestion, I think that makes sense.

Regards,
Sam

PR #38, opened 2020-11-09
Updated query notebooks

abarciauskas-bgse:
I renamed and updated the ATL08 notebook to include examples of how to use the features service and adds documentation about using the right MAAP base stack for running pdal.

I also restructured the GEDI cal/val page, building off the existing examples.


wildintellect:
TODO: Add text explaining how the feature service filters points based on zoom to reduce returns for visualization.

abarciauskas-bgse:
Closing this as stale

PR #42, opened 2021-01-12
Changed directory of Git Clone button image in git_clone_repo.ipynb

spa0002:
- Changed directory of Git Clone button image in git_clone_repo.ipynb


wildintellect:
I ran into a bug. If I try to clone into my user folder this fails. If I clone into the root of the workspace it succeeds. I think this might be due to the user folder being backed by AWS as opposed to local to the workspace.
Screenshot from 2021-01-27 15-37-18

I think this get's at a bigger issue, where we need a guide on file management that clearly diagrams what is local to the workspace vs what is cloud based and the limitations of each.

For this PR though, I think it would be greatly improved to have a few more screen shots, walking through cloning, the git tab, an opening this file (full circle). Perhaps we can find an external reference to link that provides more info on the git extension to Jupyter that is in use for further reading.

wildintellect (changes requested):
See comments in PR.

wildintellect (approved):
I think we should make a note that cloning into /Projects is the recommended location. We should test if it works elsewhere. /Project/maap-users/ fails but that's a permissions bug.

PR #43, opened 2021-01-25
Updated edav_wcs_data.ipynb

spa0002:
edav_wcs_data.ipynb has updated comments and elaboration of the code contained in cell blocks


wildintellect (approved):
Change look good. Longer term question is if we need to add more content about WCS, particularly around how to find EDAV data. We can also consider moving some of the later visualization information to a separate guide. Non of these block this PR.

wildintellect:
Just flagging, while testing, there is a [bug](https://github.com/geopython/OWSLib/issues/737) in the recent release of owslib.
Temporarily while it's being fixed, force the version `%pip install -q owslib==0.21.0`

PR #68, opened 2021-06-16
Visualize SRTM COGs from cmr-stac using MosaicJSON

slesaad:


wildintellect (approved):
Thanks for modifying the mosaic build to speed it up. I made a few modifications to the STAC query using a bbox so we get a continuous area, and some minor modifications to the zooms and simplification of the tile endpoint usage.

PR #74, opened 2021-06-30
Test gitlab trigger

abarciauskas-bgse:
Borrowed (heavily) from how https://github.com/MAAP-Project/maap-jupyter-ide is configured to trigger gitlab pipelines.

This now triggers a pipeline in gitlab and the gitlab project is also updated to run tests on whatever branch is passed as `GITHUB_REF`.

WARNING: This passes github CI just if the request to run the pipeline in gitlab succeeds. If the pipeline itself fails, that gitlab pipeline failure will not show up as an ❌ here. For now I think that's ok but might be worth looking into later.

### DIFF
--- a/.github/workflows/trigger-gitlab.yml
+++ b/.github/workflows/trigger-gitlab.yml
@@ -7,12 +7,14 @@ name: Trigger GitLab CI
 on:
   push:
     branches:
+      - master
+      - develop
       - ab/trigger-gitlab-tests
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   # This workflow contains a single job called "build"
-  build:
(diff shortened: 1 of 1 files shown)


wildintellect (approved):

wildintellect:
Agreed would be great to get the pass/fail results back on this.
Shouldn't the trigger be on PR, rather than once it's merged?

wildintellect:
so `[pull_request]` and then the REF should be the PR url?

abarciauskas-bgse:
@wildintellect I think I agree but I'm not sure I know how to implement, could you create a commit to show what you mean?

wildintellect:
Opened PR with what I'm thinking #76

PR #77, opened 2021-07-02
Made updates to notebooks where images were not showing correctly

spa0002:
- Updated markdown blocks in various ipython notebooks where images were not appearing correctly.


docs/source/search/searching_edsc_gui.ipynb
```diff
@@ -13,7 +13,7 @@
    "source": [
     "The Earthdata Search Client (EDSC) allows users to search, preview, download, and access EOSDIS Earth observation data. The MAAP has its own separate instance of NASA's EDSC which is located at [search.maap-project.org](http://search.maap-project.org). This MAAP ESDC provides a graphical user interface (GUI) for the MAAP's Common Metadata Repository (CMR) in order to make the process of discovering data easy for MAAP users. \n",
     "\n",
-    "![Earthdata Search Client GUI](../_static/edsc_gui.png)\n",
+    "![Earthdata Search Client GUI](../_static/edsc_gui.PNG)\n",
```
wildintellect, line 16: nit: `git mv` to make the file name lowercase instead, so all the ext are consistent case.

wildintellect (approved):
Makes sense to me, alt text should be screen reader compatible so no underscores is right. Only minor suggestion was to rename the PNG to png so the case is consistent in file names.

wildintellect (approved):

PR #87, opened 2021-07-13
add ogcapi features test notebook

philvarner:
Notebook for validating the integration of the ogcapi features proxy https://github.com/MAAP-Project/maap-api-nasa/pull/21


wildintellect (approved):

wildintellect:
As I mentioned in https://github.com/MAAP-Project/maap-api-nasa/pull/21#issuecomment-879429731 I think `ogc-features` or `features` is a better endpoint name.

philvarner:
> As I mentioned in [MAAP-Project/maap-api-nasa#21 (comment)](https://github.com/MAAP-Project/maap-api-nasa/pull/21#issuecomment-879429731) I think `ogc-features` or `features` is a better endpoint name.

Fixed that and updated the notebook

abarciauskas-bgse:
@wildintellect did you run this notebook in the ADE? I think we would agree that all notebooks should be executed in the ADE before merge. I think if you have run it I'm ok with merging it without running it myself. On the fence about if the notebook runs via gitlab workflow this "counts" as having run it in the ADE. Right now we need to merge #91 before this notebook can run there as a PR to develop.

wildintellect:
@abarciauskas-bgse I did not run it, won't it fail since the api endpoint isn't deployed yet?
I thought part of the CI testing was so that we have reason to think it's works in the ADE.

abarciauskas-bgse:
>I did not run it, won't it fail since the api endpoint isn't deployed yet?
right, of course! I wasn't thinking about how this is meant to be the "failing" test to drive development 🙃 

In taking a closer look, since this PR uses a locally running instance of the API for testing, I think we should hold off on until https://github.com/MAAP-Project/maap-api-nasa/pull/21 is merged and deployed. So this notebook PR I think should be a draft and we should update `api_url_root` from http://localhost:5000 to be https://api.maap-project.org. @philvarner what do you think? I agree being able to test against a locally hosted version is helpful so if you want to add a local testing option that works for me too.

PR #92, opened 2021-07-21
PR for MosaicJSON notebooks made for v2d1

spa0002:
- Adding srtm-stac-mosaic.ipynb and ade_mosaic.ipynb
- updated visualizing.rst to account for srtm-stac-mosaic and ade_mosaic notebooks
- updated querying.rst to account for ept_stores.ipynb name change


wildintellect (changes requested):
@spa0002 why are there Landsat data files in the docs directly? Those should not be in git.
PS: The example files to correct the ADE mosaic example are being copied over, will update when they are available for inclusion in this doc.

abarciauskas-bgse:
@wildintellect that was actually my suggestion to reduce confusion and issues with file paths. Open to other suggestions but chatting with Sam it sounded like there were issues (or at least I had issues) access the files which were located on the `/alexdevseed` path

wildintellect:
The issue is that files were not in GCC, they only existed in Pilot Ops - which have now been copied over. Putting them in the repo assumes users will clone the repo, which is preferred but not always the case (many will copy/paste). Also it adds 300 MB to the clone, and more if we ever change the files (as is we'd want to expunge the binary large files with BFG if we do drop them).

Alternatives:
* We could ask for a more permanent platform folder for demo data, that would always exist for MAAP users?
* We can create a much smaller dataset for embedding in the docs.

abarciauskas-bgse:
I think the approach we take, as you indicate, ideally will work regardless of if the user clones the documentation repo and also uses a small amount of storage to reduce network transfer just to run an example notebook. In this case I think it would make the most sense to copy small example tifs from a publicly available S3 storage location.

wildintellect (approved):
Let's open a ticket(s) for later:
1. Create smaller sample data and store publicly
2. Cleanup of repo history to remove the large Binaries with [BFG](https://rtyley.github.io/bfg-repo-cleaner/)

PR #105, opened 2021-12-20
Ab/access gedi l3 v2 data

abarciauskas-bgse:
Updated GEDI L3 data access example with version 2 of the dataset


wildintellect:
Looks like there are 10 granules now instead of 5. Also note a related ticket https://github.com/MAAP-Project/ZenHub/issues/495

PR #115, opened 2022-04-04
Update GEDI L3 collection identifier

abarciauskas-bgse:
The code to access GEDI L3 data fails as-is because the concept id is no longer active (from what I can tell)

I executed the updated code in ade.ops and it works as expected
Screen Shot 2022-04-04 at 4 14 57 PM

 except that I think it's downloading the data from ORNL DAAC and not earthdata cloud S3 location

Screen Shot 2022-04-04 at 4 18 14 PM

so this code should be updated so _location will be the S3 URL if it exists in the metadata

https://github.com/MAAP-Project/maap-py/blob/87efa1ea31098bc9cc43bbf04d3ddf2388caf7d2/maap/Result.py#L22

but will open a maap-py ticket for that

### DIFF
--- a/docs/source/access/accessing_external_data.ipynb
+++ b/docs/source/access/accessing_external_data.ipynb
@@ -109,8 +109,7 @@
     "# search for granule data using CMR host name, collection concept ID, and Granule UR arguments\n",
     "results = maap.searchGranule(\n",
     "    cmr_host='cmr.earthdata.nasa.gov',\n",
-    "    collection_concept_id='C2067521974-ORNL_CLOUD',\n",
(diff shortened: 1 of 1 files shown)


spa0002:
I got an error on code block 4 when I tried running the notebook:
image

It works if I change 
`download_2 = results[0].getData()` 
to
`download_2 = results[0][0].getData()`

spa0002 (commented):
[](https://user-images.githubusercontent.com/55285758/162064547-be475899-3436-4581-b965-c0e7ec687ca6.png)I got an error on code block 4 when I tried running the notebook:
image

It works if I change
download_2 = results[0].getData()
to
download_2 = results[0][0].getData()

abarciauskas-bgse:
I'm going to close this as stale.

PR #118, opened 2022-05-20
Update granules.ipynb

spa0002:
Update Searching for Granules in MAAP notebook for docs.maap-project.org/


wildintellect (changes requested):
update the MAAP api url `maap = MAAP(maap_host='api.ops.maap-project.org')`

TODO: new ticket to fix hard to use table to fix table https://docs.maap-project.org/en/latest/search/cmr_collection_table.html

wildintellect (approved):

PR #121, opened 2022-05-26
Update accessing_external_data.ipynb

spa0002:
- Updated cell 4 such that `download` variable gets the first granule from `results`.

- Updated wording to reflect changes to the code


wildintellect (commented):
NASA Goddard Commercial Cloud (GCC) MAAP ADE <- that's not the right way to name it. Internally it was important to distinguish between Ops, UAT, and DIT but not for most users. @kbugbee1 any suggestions?

 **External data access application approval**, Which permission is required for MAAP user? or does it depend on which collections they need access to?

Why are we using NASA CMR for the 2nd example? It's good to show data not in MAAP CMR but maybe we need to be clear on the example. HLS would be a good dataset to demo, GEDI L3 is odd because we do have that in our CMR (links to DAAC for download).
 
I realize @spa0002 might not be able to answer these questions. Can reassign next sprint.

sumantjha (commented):
1. docs/source/_static/page_with_authorized_application.png
It shows your name up there. Not a blocker and entirely your preference, but might be a privacy issue. May be block it up with something. 
2. docs/source/_static/profile_page.png
Same comment as above.

wildintellect (approved):

PR #134, opened 2022-11-29
Delete ept_stores.ipynb

abarciauskas-bgse:
We're not maintaining this datastore (and it's using an old version) so this documentation should be removed


spa0002 (approved):
lgtm, notebook is removed from site.

Todo: `query/ept_stores.ipynb` should be removed from querying.rst since the notebook is no longer needed in the toc.

abarciauskas-bgse:
@spa0002 good call on the querying.rst - I removed the reference there as well.

PR #135, opened 2022-11-29
Add stac to branches in workflow

abarciauskas-bgse:


abarciauskas-bgse:
I don't think this is needed

PR #158, opened 2023-03-21
Update trigger-gitlab.yml

spa0002:
Updated yml file to run job after pushes to the develop or master branch.
There was a syntax error with set-output. I do not see the error when pushing to the test-gitlab-trigger branch after updating the syntax.


wildintellect:
@sbquinlan did you already fix this?

Issue #145, opened 2023-01-18
Update Github Action before May 2023


wildintellect:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

```
echo "##[set-output name=branch;]$(echo ${GITHUB_HEAD_REF#refs/heads/})"
shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

@abarciauskas-bgse you can reassign this, but whoever fixes probably needs docs on what this job is supposed to do, and a method to test it.

abarciauskas-bgse:
@wildintellect I actually think this error is because of who+how of the PR - @marjo-luc is not a collaborator on this repo and made the PR from a fork. The documentation on the [actions secrets page](https://github.com/MAAP-Project/maap-documentation/settings/secrets/actions) says:

>Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.

So I think the fix should be: make @marjo-luc a collaborator, and if she can make a test PR which succeeds we can close this. 

I gave the MAAP-Project/maap-all team (which Marjorie is a member of) write access to this repository so I think that counts as making them all collaborators.

And just for clarification (although I think you know this, but for the sake of anyone else reading this issue), this github action only calls a gitlab job to run, so when it succeeds it just means we've successfully triggered the job, but the job likely fails.

abarciauskas-bgse:
The GITLAB_TRIGGER_URL will be at https://mas.maap-project.org/root/maap-tests (not sure if that's the exact path) and I know I created the access token under my username in that UI (although I'm not exactly sure where at the moment since Settings > Access Tokens says "This project has no active access tokens.")

wildintellect:
@abarciauskas-bgse that does solve the mystery of what threw the error but does not actually fix the github action feature that is changing (which is the actual topic of this ticket).

abarciauskas-bgse:
@wildintellect I see, my mistake! Then yes we do need to change our actions yaml file according to the [blog](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

PR #160, opened 2023-03-28
Replace 'set-output'

sbquinlan:
Fix MAAP-Project/maap-documentation#145 and NASA-IMPACT/active-maap-sprint#411


sbquinlan:
The post to CI is 404. Not sure why without access to the secrets, but from the action run it looks like my update is correct. Requesting review to confirm.

abarciauskas-bgse (approved):

PR #164, opened 2023-04-04
Feature/add devfile info

grallewellyn:
The create workspace notebook previously said: "Name your workspace, choose your stack and the amount of RAM you will need." and I think it is possible that was created at a time there was a different workflow because I am not seeing a way to specify the amount of RAM that we want when creating a workspace and other screenshots look different in read the docs. It is also not clear exactly where we should edit in the devfile, so I added that information. 
I also don't see a way to add a git project or specify the organization, so I took this out of the description. Maybe this is something else that has been taken out with newer versions. Correct me if I am wrong. 

Screenshot 2023-04-04 at 1 13 04 PM


wildintellect:
I'm going to just note agreement with @grallewellyn the ram selection was different in V1. Git projects can still be added, but I'm not sure it's recommended over just doing a git clone was the workspace is up.

@grallewellyn was this tested on DIT, or Ops?

grallewellyn:
@wildintellect This was tested in both DIT and OPS and both appear to be creating a workspace the same way

rtapella (approved):
lgtm

PR #144, opened 2023-01-10
added doc. for new jobs ui

marjo-luc:
Initial commit for new Jobs UI documentation. This is a new UI that gives users a view of the job processing they have done and the ability to submit jobs. They will be able to view job metadata such as products generated, errors produced, and job status. The UI is accessible from users' jupyter workspaces.

PR #165, opened 2023-04-04
V2 merge and tag

wildintellect:
This is rolling up all the recent documentation updates into the master before we tag V2 and then re-organize.


wildintellect:
A couple of issues
- The Job UI work by @marjo-luc applies to the new "V3" job ui. <- this just highlights the need for a better way to manage future version docs.
- The requirements file has a `~=0.3.x` for pystac, which is already and old version (not a breaking issue, but a todo)

Thoughts on how to handle @rtapella ?

docs/source/platform_technical_documentation/ade_custom_extensions/jobsui.ipynb
```diff
@@ -0,0 +1,89 @@
+{
```
sumantjha, line 1: <p>View and Submit jobs icon is not visible in my version of ADE. Is it still there on other's version or has it been removed. In that case, this page needs to be modified. However, since we are going to make this version of RTD v2, this issue can also be fixed in the V3 version of RTD.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/165/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

sumantjha: Looks good to me. Ready to merge.

wildintellect: I think you'd only see this in the DIT version of the ADE. One option is we could revert https://github.com/MAAP-Project/maap-documentation/pull/144 , and reapply it after V2 tagging?

sumantjha: Yes, reverting and reapplying should resolve this.

rtapella:
Grace mentioned that `feature/add-devfile-info` fixes some documentation that is already in Ops (i.e., should be v2)

rtapella:
> * The Job UI work by @marjo-luc applies to the new "V3" job ui. <- this just highlights the need for a better way to manage future version docs.

This is isolated in the docs and I think it's okay to include here.

PR #166, opened 2023-04-05
202304 reorg

rtapella:
Reorganization at top level


docs/source/index.rst
```diff
@@ -9,15 +9,10 @@ Welcome to the MAAP User Documentation!
...
-  intro.rst
-  searching.rst
-  visualizing.rst 
-  accessing.rst
-  querying.rst
-  userdata.rst
+  getting_started.rst
+  science_examples.rst
   tutorials.rst
-  faqs.rst
```
wildintellect, line 19: FAQs is delisted here, but the FAQ folder still exists. Is this intentional to hide pages (orphan)? More generally should the folder and section names match?

docs/source/science_examples.rst
```diff
@@ -0,0 +1,16 @@
+Science Examples
+=======================================
+
+.. toctree::
+  :maxdepth: 2
+  :caption: Science Examples:
+
+
+  tutorials/HLS/HLS_search.ipynb
```
wildintellect, line 9: Same question from before, do we need to update the folder name to science_examples, or at least `science` since this is reflected in the urls? I don't think we're worried about breaking links at this point, so a url change should be ok.

docs/source/system_reference_guide/ade_custom_extensions.rst
```diff
@@ -0,0 +1,10 @@
+ADE Custom Extensions
+=======================================
+
```
wildintellect, line 3: There's no text explaining what is in this section.

docs/source/science_examples.rst
rtapella: Yes this is a good time to change the naming conventions etc.

The folder names don't need to match the rst.  I seem to recall leaving things alone as much as possible.

rtapella: Speaking of which... how do we feel about `Science Examples` ?

docs/source/index.rst
rtapella: I moved the FAQ into the System Reference Guide section. Maybe I forgot to move the `faq/` folder along with `faqs.rst`

docs/source/system_reference_guide/ade_custom_extensions.rst
rtapella: True!  There is also no explanation on the existing page: https://docs.maap-project.org/en/develop/ade_custom_extensions.html

rtapella: (I am not opposed to adding a description as part of the re-org)

wildintellect (commented):
Overall my only concern before approving this is that the folder names and heirarchy don't seem to match the Titles of the sections which is going to lead to some confusing urls, and complications for editors. Some examples were pointed out inline comments.

docs/source/science_examples.rst
wildintellect: They don't need to match 100% but this is confusing because a different section has Tutorials in it's name. `science examples` is fine for now, I would make the folder `science` to make a nicer url.

docs/source/index.rst
wildintellect: yup that's it. so maybe just a move is in order, since these are clearly FAQs only for System section.

docs/source/system_reference_guide/ade_custom_extensions.rst
wildintellect: Let just open a ticket to go through and add Descriptions to all the landing pages, and not do it in this PR.

docs/source/index.rst
rtapella: Oh I think I left it at the top level because all the image-links have to be altered for the faq files to work. I'll move it and update them.

(This is why I was asking about the `_static` folder and whether I can make new ones to move entire sub-trees of files around.)

rtapella: fixed in commit: 4854358c3f32ec484b1dc25ae356b734e88b6d0f

rtapella: I also have a subsequent commit to rename faqs.rst to faq.rst for consistency

wildintellect (approved):

docs/source/system_reference_guide/ade_custom_extensions.rst
rtapella: see: https://github.com/MAAP-Project/maap-documentation/issues/168

docs/source/science_examples.rst
rtapella: commit: 88ba9051f4c0be68f9c405affad6243c9f6e705c

renamed `tutorials/` to `science/`

rtapella:
I have also changed `intro/` to `getting_started/`

wildintellect (approved):

PR #167, opened 2023-04-05
Feature/devfile disclaimer

grallewellyn:
Added the following disclaimer to create workspaces as per @gchang 's suggestion 

**DISCLAIMER:** MAAP might change the way that we do workspace memory limits in the future, including restricting the memory that a user can allocate for their workspace.


wildintellect:
@grallewellyn should we mention a sensible upper limit, and put a line "As a courtesy to other users only increase this limit if your kernel keeps running out of memory and crashing. If you're not sure why your kernel is crashing please contact the development team"

PS: This won't be merged until after #166

grallewellyn:
@wildintellect I agree, but what if we change it slightly to "As a courtesy to other users only increase this limit if your kernel keeps running out of memory and crashing. If you're not sure why your kernel is crashing **after increasing your workspace's memory limit,** please contact the development team"

Bold is what I am suggesting to add

wildintellect:
@grallewellyn what's the upper limit? I had suggested to some users not to go over 16 GB without asking.

grallewellyn:
@gchang has mentioned 32GB, but maybe he can provide more clarification. I am not sure if this is something we want to decide on now and add to this disclaimer, or leave the disclaimer more vague. 

@wildintellect

wildintellect:
I'd prefer we set a suggestion at minimum, which we can always change. Reason - users will go as big as they can (also there is a physical limit to the node).

@grallewellyn can you also `rebase` or merge `develop` into your branch to bring it up to date with our recent re-organization of the dir structure?

grallewellyn:
@wildintellect Done. 

I agree we should suggest a minimum. Should we be suggesting 16GB in the disclaimer for now? I tried changing the devfile to `16GB` and `16Gb` and the workspace fails to start. Should we be telling users to enter 15258Mi for memory?
Screenshot 2023-04-10 at 8 06 06 AM

grallewellyn:
@wildintellect @rtapella Does this disclaimer look good?

wildintellect (approved):

PR #170, opened 2023-04-08
Added ATL08 access and explore instruction. Need to figure out why it…

sumantjha:
Added ATL08 access and explore instruction. Need to figure out why it started giving error at the end all of a sudden. 
Also @wildintellect : While sending out the commit, it mentioned that there is a conflict now with the branch. I believe it is because of the changes we made while doing the restructuring with Rob.


wildintellect:
@sumantjha reports the notebook is currently broken but was working before
```
OSError: Unable to open file (unable to open file: name = 's3://nsidc-cumulus-prod-protected/ATLAS/ATL08/005/2021/11/14/ATL08_20211114213015_08161305_005_01.h5', errno = 2, error message = 'No such file or directory', flags = 0, o_flags = 0)
```

sumantjha:
If the goal is to get the data out in the geodataframe, why can't we use the same approach as ATL03: https://docs.maap-project.org/en/latest/science/ATL03/IceSat_ATL03_Tutorials.html#Show-the-subset-data-in-a-dataframe.

wildintellect:
@sumantjha you could do it that way, I forgot to link that as options 3. If you want to make that change we could include it in this PR.

emileten:
Checking the status of this -- should it be merged or closed?

wildintellect:
@emileten needs to be reassigned. @smk0033

smk0033:
WAs NSIDC ever contacted about credentials? I haven't been able to get AWS access by Sumant's way in the notebook or through the way the HLS notebook gets credentials

wildintellect:
@smk0033 I can't recall, it's a good question to raise with @freitagb and verify the functionality with @bsatoriu

CC: @abarciauskas-bgse

smk0033:
Pushed updates so far. Still needs work, but I've done some reorganizing and added the data access from Jamison's direct access notebook. I saw where you can convert from xarray into pandas, so I was going to maybe try something like that and do a visualization like ATL03? Still need to rename the doc as well and I was probably going to create one folder for ICESat/ATL notebooks like how GEDI is

smk0033:
closing to open in new PR

PR #171, opened 2023-04-10
First Iteration of V3 with new Organization

wildintellect:
Mostly this is a re-organization of all the files. Can be seen at https://docs.maap-project.org/en/develop/
The current master has been tagged https://github.com/MAAP-Project/maap-documentation/releases/tag/2.0.0 is currently https://docs.maap-project.org/en/stable/ (in addition to `latest`) and will always be accessible by it's tag in the future.

We want to merge this ASAP so we can announce new documentation on STAC.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -21,7 +21,7 @@ def setup(app):
 # -- Project information -----------------------------------------------------
 
 project = 'maap-docs'
-copyright = '2020, NASA MAAP Team'
+copyright = '2020-2023, NASA MAAP Team'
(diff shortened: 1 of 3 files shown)


abarciauskas-bgse (commented):
Thanks @wildintellect good to know about the 2.0.0 tag so we can always revisit the way things were 💭

wildintellect:
This Branch will get a 3.0.0 once it's in too, so that we can start versioning.

rtapella (approved):
small fixes have been implemented. lgtm

PR #175, opened 2023-04-10
Re-org search docs

abarciauskas-bgse:
https://docs.maap-project.org/en/stac/technical_tutorials/searching.html

This is a PR to fix the search organization. Changes:

- The default searching.rst has the TOC moved up, I just have a few sentences there about the two catalogs and a link to a new page catalog.rst
- catalog.rst includes the information about the two catalogs and how to migrate code
- I also retitled the collections.ipynb and granules.ipynb so that it's clear they use NASA's Operational CMR + maap-py
- Alex M asked a good q: "How do we tie in data access? is maap-py still the right way ( I think it is for non cloud optimized data in MAAP buckets)?"
    - I think we should create a new ticket about documentation on data access. maap-py is still a good way to access data from all of Earthdata because it includes the federated access token. But it doesn't allow for access in a  cloud-optimized way AFAIK via something like rioxarray, so we probably want to document that

### DIFF
--- a/docs/source/technical_tutorials/search/catalog.rst
+++ b/docs/source/technical_tutorials/search/catalog.rst
@@ -8,7 +8,7 @@ MAAP users are advised to use two catalogs:
 
 The https://cmr.maap-project.org catalog will be deprecated by **May 1, 2023**. Users should request collections they need from this catalog to be made discoverable in the MAAP STAC or NASA's Operational CMR if they're not already there.
 
-More information on each catalog and migrating from MAAP's CMR is detailed below.
+More information on each catalog and migrating from MAAP's CMR is detailed in the bottom of this page.
 
 =======================================
 MAAP STAC
--- a/docs/source/technical_tutorials/search/catalog.rst
+++ b/docs/source/technical_tutorials/search/catalog.rst
@@ -3,12 +3,72 @@ MAAP's Dual Catalog
 
 MAAP users are advised to use two catalogs:
 
-1. NASA's Operational CMR: https://cmr.earthdata.nasa.gov
-2. MAAP STAC: https://stac.maap-project.org
+1. Use NASA's Operational CMR to discover NASA-produced and curated data: https://cmr.earthdata.nasa.gov
+2. Use MAAP STAC for data not found in NASA CMR, and data produced by MAAP users: https://stac.maap-project.org
 
 The https://cmr.maap-project.org catalog will be deprecated by **May 1, 2023**. Users should request collections they need from this catalog to be made discoverable in the MAAP STAC or NASA's Operational CMR if they're not already there.
 
 More information on each catalog and migrating from MAAP's CMR is detailed below.
+
+=======================================
+MAAP STAC
+=======================================
+
+MAAP STAC (https://stac.maap-project.org) is dedicated to datasets not accessible via NASA's CMR, such as GEDI Cal/Val datasets, ESA datasets, and user-shared data products.
+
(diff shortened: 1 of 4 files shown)


docs/source/technical_tutorials/search/catalog.rst
wildintellect, line 1: I think Migration should be the last section of this page. After the docs about the 2 search interfaces.

docs/source/technical_tutorials/searching.rst
```diff
@@ -2,147 +2,15 @@
 Search
 =======================================
 
-MAAP's Dual Catalog
```
wildintellect, line 5: I think it's fine for this whole section on Dual Catalog to stay on the Search landing page.

docs/source/technical_tutorials/searching.rst
```diff
@@ -2,147 +2,15 @@
...
-
-CMR Access
----------------------------------------
-
-For all NASA MAAP users, access to NASA'S Operational data is provided via a federated access token. Anything that is in NASA's Operational CMR should be accessed via maap-py so that the federated access token can be used. Users can also access data from LPDAAC (and possibly other DAACs in the future) without maap-py since the workspace should have access via a role-based bucket policy on the LPDAAC cloud bucket.
-
-Anyone can access data through Earthdata Login as well.
-
-Find more documentation about how to access data in CMR in the `Access <accessing.html>`_ section of this documentation.
+MAAP provides access to data in NASA's Operational CMR (Common Metadata Repository, https://cmr.earthdata.nasa.gov) and a MAAP-managed STAC catalog (https://stac.maap-project.org).
```
wildintellect, line 5: This should probably still be a list, and maybe a short sentence/description for each, so users know which one to go to?

NASA CMR - for NASA produced and curated data.
MAAP CMR - for data not found in NASA CMR, and data produced by MAAP users.

wildintellect (commented):
A few clarifications before approving.

docs/source/technical_tutorials/searching.rst
abarciauskas-bgse: I think the issue with having the dual catalog content on the searching.rst page is then only those headers (the ones in the dual catalog documentation show up, not the other pages like "search using EDSC". I think we want the other pages to show up in the left hand navigation, so I'm going to keep the dual catalog content on a separate page. But I have modified the text in the top section of searching.rst to be a bit more descriptive as you suggested in the other comment on this PR.

docs/source/technical_tutorials/searching.rst
```diff
@@ -2,147 +2,20 @@
...
-
-Anyone can access data through Earthdata Login as well.
-
-Find more documentation about how to access data in CMR in the `Access <accessing.html>`_ section of this documentation.
-
+More information on each catalog and migrating from MAAP's CMR here: `MAAP's Dual Catalog </search/catalog.html>`_. 
 
 .. toctree::
   :maxdepth: 2
   :caption: Search:
```
wildintellect, line 16: Search Topics

docs/source/technical_tutorials/searching.rst
```diff
@@ -2,147 +2,20 @@
...
   
+  search/catalog.rst
```
wildintellect, line 18: Cleanup Chore Remove: 
search/cmr_collection_table.ipynb
search/deprecated_collection_table.ipynb

wildintellect (approved):

Issue #186, opened 2023-05-04
ReadTheDocs setup.py build fail


wildintellect:
Our config is using an old deprecated method to build, as well as an old python version (3.7). On May 3 this started resulting in build errors.

```
 python -m pip install --exists-action=w --no-cache-dir -r requirements.txt
 Requirement already satisfied: sphinx>=1.4 in /home/docs/checkouts/readthedocs.org/user_builds/maap-project/envs/getting-started-pr-branch/lib/python3.7/site-packages (from -r requirements.txt (line 1)) (1.8.6)
Collecting pandoc (from -r requirements.txt (line 2))
  Downloading pandoc-2.3.tar.gz (33 kB)
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [1 lines of output]
      ERROR: Can not execute `setup.py` since setuptools is not available in the build environment.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
```
This does not seem to be caused by any change to our code. So best action is to upgrade using a config file https://docs.readthedocs.io/en/stable/config-file/v2.html

wildintellect:
https://github.com/readthedocs/readthedocs.org/issues/8775

wildintellect:
Disabling system_packages results in a different error
```
Running Sphinx v1.8.6
loading translations [en]... done

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/maap-project/envs/latest/lib/python3.7/site-packages/sphinx/registry.py", line 472, in load_extension
    mod = __import__(extname, None, None, ['setup'])
  File "/home/docs/checkouts/readthedocs.org/user_builds/maap-project/envs/latest/lib/python3.7/site-packages/sphinx/builders/linkcheck.py", line 19, in <module>
    from requests.exceptions import HTTPError
  File "/home/docs/checkouts/readthedocs.org/user_builds/maap-project/envs/latest/lib/python3.7/site-packages/requests/__init__.py", line 43, in <module>
    import urllib3
[...]

PR #184, opened 2023-05-01
Getting started overhaul

rtapella:
Built preview: 
https://docs.maap-project.org/en/getting-started-pr-branch/getting_started.html

Overhaul of all the Getting Started docs. This is a combo of:
- the old GS docs
- Alexey S's Quickstart Guide
- new content and rewritten content
- Emile's rewrite of the conda environment stuff

If you have time please go through the whole thing and try to follow the instructions. Some areas that need review:
- About and the new overview docs
- Jupyter interface overview section of "setting up your account and workspace"
- MAAP storage options section of "setting up your account and workspace"
-  Using maap.py from a Python notebook needs some content to help get started in "Writing code"
- Some helpful starter templates in "Writing code" (Sujen?)
- working with GitLab and GitHub in "Writing code"
- clean up the conda section and replace with Emile's new page if that makes more sense (delete content and reference the new section)
- Registering an algorithm needs testing. Should be able to: 1. clone a demo algorithm 2. edit it 3. register it.
- Run and monitor a job: update with new jobs UI

PR #187, opened 2023-05-04
Fix build configuration that broke

wildintellect:
Solving #186 
Not really sure why it broke, something in the underlying image ReadTheDocs was using might have changed.

Updated to use .readthedocs.yml which overrides some Advanced Settings in the web interface.
Also bumped the python version to 3.11 from 3.7 and versions of several packages.

This fix will need to be backported to all active branches before new PRs can be accepted or branches built. Starting with #184 that's currently broken.


grallewellyn:
I am able to successfully run 
```
cd docs
make html
cd build/html
python3 -m http.server
```
and see the version of read the docs before we made the changes to getting started. 
But I am not sure if this problem was for local builds too

rtapella:
> I am able to successfully run
> 
> ```
> cd docs
> make html
> cd build/html
> python3 -m http.server
> ```
> 
> and see the version of read the docs before we made the changes to getting started. But I am not sure if this problem was for local builds too

Local builds were working for me, too, but the CI was (is) not.

wildintellect:
To be clear only ReadTheDoc builds were impacted, but it's good that these changes did not break your local (although for true test you probably would need to reinstall python libs from the requirements into a fresh environment)

grallewellyn:
I tested with a fresh environment created by the current requirements.txt and that worked

rtapella (approved):
The fix works!

PR #200, opened 2023-05-18
Changed tutorial names

smk0033:
Changed the science tutorial names. Ignore all of the other changes made to HLS for now, I started working on other tickets with that one. I also adjusted some of the file names


wildintellect (changes requested):
GEDI_02B and GEDI L2B are the same thing. For consistency we need to have the same naming style (I prefer the L2A L2B style).

- Do the notebooks really need tutorial in their name?
- What's the difference between the 02B and L2B notebooks?

As noted in slack, the HLS notebook changes should be removed from this branch before we can approve as it's not part of the task/ticket.

jjfrench:
@wildintellect, looks like Aimee's (02B) is a tutorial on using the STAC catalog while Nathan's (L2B) is using NASA CMR search

wildintellect:
Are we keeping 02B in MAAP STAC @abarciauskas-bgse? I think we removed it as a duplicate of NASA CMR, so delist or delete the notebook?

rtapella (approved):
Superficially this is fine. I don’t know enough about the details to comment more.

smk0033:
@wildintellect I'll switch over the GEDI tutorials to use L instead of 0! As for the two L2B ones, is the CMR one going to be deleted as you suggested above? If not, I'll have to add something to differentiate the two (Like GEDI_L2B_stac and GEDI_L2B_cmr?)

smk0033:
I'll also go ahead and remove "tutorial" from the files names so they'll just be the datasets

smk0033:
New commit. Left GEDI_02B as-is in case it does get deleted, if not then I'll have to readjust the names to separate the GEDI_L2Bs. I also think I fixed the HLS file?

docs/source/science/GEDI/GEDI_02B.ipynb
wildintellect, line 1: Should we hide this from the Navigation if it doesn't actually work anymore?

docs/source/science/GEDI/GEDI_L4A.ipynb
wildintellect, line 1: Change Plot to Visualize to be consistent with other notebooks?

docs/source/science/GEDI/GEDI_L4B.ipynb
wildintellect, line 1: This is not a blocker on this PR, but this tutorial should be updated to use L4B granules directly from Earth Data Cloud, it should not 'download'.

docs/source/science/HLS/HLSL30.ipynb
wildintellect, line 1: I'm not sure Visualize is the correct term for this title - Composite the original term is much closer to the science term to take mulitple rasters over the same place and combine them. In this case many files are being combined to find the max NDVI for each pixel over the time period selected.

docs/source/science/GEDI/GEDI_02B.ipynb
smk0033: That sounds good! Is there a way to do that using git in the terminal, or is that something we will do through Github itself?

wildintellect: The navigation is controlled by https://github.com/MAAP-Project/maap-documentation/blob/develop/docs/source/science_examples.rst just remove the line referencing the notebook and commit the change.

docs/source/science/GEDI/GEDI_L4B.ipynb
smk0033: That is noted! Since it's not a blocker on this PR, does it need to go ahead and be updated, or can I update when I go through the science tutorials and change the CMR/ops stuff?

wildintellect: I would make a ticket specific for updating the notebook. This would correspond to when you review the notebooks for template consistency and will be working with 1 notebook at a time.

smk0033:
Also updated GEDI_L4B to "Search and Visualize"

wildintellect (approved):

PR #209, opened 2023-05-26
Afrisar name updates

smk0033:
Maybe the third time is the charm?

I used git mv to rename the file and folder, and that seemed to work. I think my issues are coming in when I actually go into the file to update the tutorial name itself. I don't remember it giving me such a hard time when I renamed the other science tutorials, and it tells me the file has been "modified" when I do git status - is github reading it in as a whole new file still?


wildintellect (approved):

wildintellect (changes requested):
What about the update to the science_examples.rst?

wildintellect (approved):

PR #218, opened 2023-06-06
Broken Science Tutorial Fixes

smk0033:
Updating science tutorials that are broken. Reference ticket: https://github.com/orgs/NASA-IMPACT/projects/29?pane=issue&itemId=27480247 (issues are under each individual ticket). Wanted to go through and fix errors to make sure notebooks were fully functional before going in and updating per new guidelines.

Fixes so far:
- NISAR tutorial: updated paths
- GEDI_L2B: updated CMR host and removed a search limit, otherwise there was an "invalid token" error when searching for the granule

Will commit other science tutorials as they get fixed, and will create a separate PR for technical tutorial fixes after.


wildintellect (commented):
* Does the Limit really cause the token error (the old CMR info should be the issue). Can you try adding that back to the `searchGranule`
* For the NISAR tutorial we should probably use the "best practices" and have a user download to a `data` directory that they make before downloading.  See https://docs.maap-project.org/en/latest/technical_tutorials/access/accessing_data.html (Cell 4)

smk0033:
Sorry, definitely meant GEDI_L2B tutorial in the commit above

smk0033:
@wildintellect I created a new data directory for the .h5 file that downloads. Should the HH files also go into that new directory? I wasn't quite sure on how to change the path for those if so, so they still just download into the NISAR file

smk0033:
Tutorials updated so far: NISAR, GEDI_L2B and GEDI_L4B.
GEDI_L4A is in progress.
@wildintellect since the GEDI_02A/L2A dataset was removed from STAC, should we use another collection or convert to a CMR tutorial?

smk0033:
I think this PR is fully ready for review. I'll probably put GEDI_L2B in it's own PR since I'll have to use a new collection

omshinde:
Thanks @smk0033 
I am working on reviewing this PR. 
CC: @wildintellect @jjfrench

docs/source/science/NISAR/NISAR_access.ipynb
```diff
@@ -2,23 +2,23 @@
  "cells": [
```
jjfrench, line 2: <pre class="ql-syntax" spellcheck="false">Line #2.    # ! pip install h5glance
</pre><p>Can probably be removed since it's listed above</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

smk0033: <p>I think I added this because it says glob cannot be found?</p>

jjfrench: I'm guessing it's erroring out and not continuing the package installs - can probably remove glob from the installs or install glob2?

smk0033: <p>Looks good! I just replaced it was glob2. The only other one giving me issues is osgeo, it says there is a problem with the package and not pip. Says to use gdal instead?</p>

omshinde (commented):
Added some comments in the notebook discussions.

docs/source/science/NISAR/NISAR_access.ipynb
jjfrench, line 2: <pre class="ql-syntax" spellcheck="false">Line #2.    from osgeo import gdal
</pre><p>Might fail if we're no longer installing osgeo</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

smk0033: <pre class="ql-syntax" spellcheck="false">Line #2.    from osgeo import gdal
</pre><p>Hmm, it actually still seemed to work when I ran through it. Should I just update it to import gdal like in the beginning of the tutorial? Or can that be removed, along with numpy since they were installed in the beginning?</p>

jjfrench: We can keep the imports down below since that's the only place they're being used

I think that part of the problem is that we're installing using "!" instead of "%" so it's could be different environments? Maybe we switch this tutorial to use something like rasterio?

jjfrench: oh, I see - I don't know if we need to import numpy up above if it's not being used until down below

jjfrench:
```
import numpy as np
import rasterio
import matplotlib.pyplot as plt


# Extract a subset of the SLC to display
x0 = 0
y0 = 10
x_offset = 1000
y_offset = 1000

with rasterio.open('HH.slc') as ds:
    # Define the window of data to read
    window = rasterio.windows.Window(x0, y0, x_offset, y_offset)
    
    # Read the data from the specified window
    slc = ds.read(1, window=window)
    print(slc)

fig = plt.figure(figsize=(14, 12))

# Display amplitude of the slc
ax = fig.add_subplot(2,1,1)
ax.imshow(np.abs(slc), vmin = -2, vmax=2, cmap='gray')
ax.set_title("amplitude")

# Display phase of the slc
ax = fig.add_subplot(2,1,2)
ax.imshow(np.angle(slc))
ax.set_title("phase")

plt.show()
```

Something like this, and including rasterio in the installs above when using "%"

smk0033:
<p>Good catch, I definitely didn't notice that. Thanks!!</p>

---

 <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/#comment-1599431104'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

smk0033:
<p>I'll test it with api.maap-project.org again, it may have originally not worked due to that bug (or maybe I had gotten it confused by accident) so I'll update it if it runs using api.maap-project.org</p>

---

 <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/#comment-1599431082'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

docs/source/science/NISAR/NISAR_access.ipynb
smk0033: <p>Okay, it should be updated now </p>

jjfrench:
<p>For maap.profile.account_info()["username"]: api.maap-project.org for ade.maap-project.org, api.ops.maap-project.org for ade.ops.maap-project.org</p><p> </p><p>They don't have job submissions setup for api.maap-project.org yet though, you'll only get 500's</p>

---

 <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/#comment-1599436361'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

jjfrench:
<p>since Anish hardcoded his username into the AOI path, I think we should use the api.ops.maap-project.org host for now</p>

---

 <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/#comment-1599436361'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

smk0033:
<p>Yeah, sometimes it works and sometimes it doesn't, at least with the .ops in it</p>

---

 <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/#comment-1599436361'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

docs/source/science/GEDI/GEDI_L4B.ipynb
```diff
@@ -2,15 +2,15 @@
  "cells": [
```
wildintellect, line 2: <p>The previous code showed how to find the collection shortname. We should either drop this section (and link to an example of How to find a Collection name), or revert back.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

smk0033: <p>The previous way didn't work for some reason, not really sure why but it wouldn't pull that collection even though the short name was correct. There was also something wrong with pprint and it wouldn't print it, so that's why we reverted to this way of pulling with the short name</p>

docs/source/science/NISAR/NISAR_access.ipynb
jjfrench, line 2: <p>I guess the location will depend on where you're running this notebook? Should we try to make this relative? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='21'/>

jjfrench, line 2: <p>same comment as above - Should we try to make this relative?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/218/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

jjfrench: Right, it would technically not have to be within the repo directory, just the data directory

omshinde:
Thanks for the changes @smk0033 and @jjfrench ..Although, I am late to the party, it looks good to me as well! :)

Issue #251, opened 2023-06-27
Confusion over installing conda packages from a yaml to an existing env


wildintellect:
Creating a new Conda env from an env file is straight forward and well documented with https://docs.maap-project.org/en/develop/system_reference_guide/custom-environments.html#Basic-custom-environment

However https://docs.maap-project.org/en/develop/system_reference_guide/custom-environments.html#Extending-the-base-environment-in-a-given-workspace-session. Needs an additional section on how to update an existing env from a yaml defining packages. (Also remove the period on the heading)

Example:
```
mamba env update -n base --file above_env.yml
```
based on info from https://stackoverflow.com/questions/42352841/how-to-update-an-existing-conda-environment-with-a-yml-file

As reported by @pahbs in Slack on Jun 27, 2023.

emileten:
@wildintellect we have that already https://docs.maap-project.org/en/latest/system_reference_guide/custom-environments.html#Updating-an-existing-environment-with-a-configuration-file

However the fact that you didn't find it means something is wrong. Sounds like it should go in https://docs.maap-project.org/en/latest/system_reference_guide/custom-environments.html#Extending-the-base-environment-in-a-given-workspace-session, but I had explicitly avoided mentioning configuration files in that first section, to leave the config files related topics to the next part.

wildintellect:
I think it's easy to miss because it's in a paragraph. We should call it out in a code block with a title or caption.

emileten:
addressed in https://github.com/MAAP-Project/maap-documentation/pull/252

PR #252, opened 2023-06-28
add missing reference to example yaml files notebook, and clarify section about env update based on a config file

emileten:
## Clarifies the documentation explaining how to update environments

See https://github.com/MAAP-Project/maap-documentation/issues/251

## Adds a reference to the page with the example config files

In https://github.com/MAAP-Project/maap-documentation/pull/197/files I had changed the way we were pointing to example conda configuration files. The intent was to have a notebook showing the content of these files and having this notebook as a separate page of the docs, with the explanation page pointing to the latter page for examples. 

I had forgotten to reference that examples page, so it's not showing up in the docs. 

Fixing that here. See my screenshot.

Screenshot 2023-06-28 at 12 13 38 PM


docs/source/system_reference_guide/custom-environments.ipynb
```diff
@@ -49,14 +49,19 @@
     "\n",
```
wildintellect, line 49: <pre class="ql-syntax" spellcheck="false">This is too vague <code>mamba env update ARGUMENTS</code> 
</pre><p>users won't know what arguments. Reasonably that is actually conda/mamba docs but most people aren't going there (we should link to them).</p><p>So I think moving </p><h3>Updating an existing environment with a configuration file</h3><p>up to here actually does make sense.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/252/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

emileten: Agreed. 

Instead of moving like you suggested, replaced it with 

```
mamba env update -f config.yml
```

and then pointed to the next sections for details on what that config file could look like.

emileten:
@wildintellect : 

- made an internal link to point to internal sections of the page
- I also replaced the quite messy approach of pointing to a notebook that prints yaml files. I removed that notebook, and instead I added links to download the yaml files themselves. This way, users don't have to copy paste from the notebook printed cells...

wildintellect (approved):
I'm ok with these changes as an improvement over the current. However I have questions about finding a better way to surface the yaml examples. Option include putting the yaml in code blocks and adding a copy button #256 , pushing users to clone the repo (or automatically including it in workspaces), redirecting users to the github repo folder, or the current where it just opens in the browser window for copy/paste.

PR #257, opened 2023-07-04
Add ESA CCI V4 notebook

emileten:
Adding a notebook that demonstrates how to access and visualize the ESA CCI V4 collection from the MAAP STAC catalog. [This follows a user question about how to access this data.](https://github.com/MAAP-Project/Community/issues/709#issuecomment-1618510962).

In this notebook I turn a collection search into an array using `stackstac` and plot a time series of the average AGB over a single tile. I also show how to access the catalog field that contains the storage location information. 

@nehahunka do you think this would be useful ? I am wondering if a different, or richer, analysis would be more helpful. 

@wildintellect tagging you to get your opinion about whether this is useful. Note that this is a draft PR, there are definitely things to improve, notably standardize the notebook to our standards in terms of format and documentation, and if possible avoid requiring to create an extra environment (somehow for now I couldn't make it work without doing this).


wildintellect:
@emileten I think it's useful, would suggest adding a visualization of maybe 2 time periods side by side for some portion of the cube.

smk0033 (changes requested):
Looks good overall! I added some reviews in ReviewNB - let me know if you can't see them. Mostly just dealing with additional information in the beginning (at the very least, I'd add author(s), date and change "What are we doing in this notebook?" to "Description" and remove it as a header). You can add other sections I mentioned in ReviewNB if you wish, but I can also go in and add them a little later on. From there, I'd also add some other headers to help generalize sections

nehahunka:
Hi Emile,

I agree with your suggestion to “set up a notebook example to discover and
read this data on a notebook in the ADE and publish this notebook as a page
of the MAAP docs”. A few things - (1) it would be great if the time-series
for an arbitrary vector polygon provided by the user is used as a boundary
for calculating the stats, and (2) please add a line where some values can
be "masked out" from the map (e.g. all values below 10 are ignored). I
think that should make a nice example. However, I wonder if you should
check with Aimee first...because I am almost certain I remember her showing
us a notebook with stats for biomass being displayed in some user-friendly
graphics... Thank you!!

Neha

emileten:
Thanks @nehahunka ! What you mentioned should be easy to add. As per the notebook you mentioned, I asked @abarciauskas-bgse !

emileten:
https://github.com/orgs/NASA-IMPACT/projects/29/views/6?pane=issue&itemId=30816540

smk0033 (approved):

PR #219, opened 2023-06-07
HLS VEDA guidelines

smk0033:
Link to ticket: https://github.com/orgs/NASA-IMPACT/projects/29?pane=issue&itemId=27479193

Updated the HLS tutorial and followed [these guidelines](https://nasa-impact.github.io/veda-docs/contributing/docs-and-notebooks.html).

Changes mainly consisted of adding rendering information, a link to run the notebook in MyBinder, an "About the Data" section, and additional resources.

Open to any feedback!

PR #258, opened 2023-07-05
GEDI L2B VEDA Guidelines

smk0033:
In reference to: https://github.com/orgs/NASA-IMPACT/projects/29/views/6?pane=issue&itemId=32193528

Added rendering info, new sections (Run This Notebook, About the Data and Additional Resources), and added headers throughout the tutorial since they were missing.


jjfrench (approved):
Looks good!

emileten:
super minor suggestion unrelated to the PR content in itself @smk0033, in case you had not noticed. If you link the related issue through the project board, and the issue gets moved from the board, I believe the link gets broken. E.g. the issue link in this PR is broken https://github.com/MAAP-Project/maap-documentation/pull/219.

So maybe to ensure the link is preserved for anyone looking at the PR in the future, either of the following ? could be better (which all amount to directly link to the issue rather than the board it was added to)

- https://github.com/MAAP-Project/maap-documentation/issues/243 
- #243 
- or better, in any issue on the right there is a button to link a pull request. Could use that as well !

smk0033:
Thanks, @emileten! I had no idea those links were breaking😅 thanks for pointing that out and recommending fixes for them!

Issue #256, opened 2023-06-30
Make it easier to copy code blocks


wildintellect:
While reviewing ticket #252  we debated the best way for users to get examples of conda package yaml files.
I realized that a copy button on code blocks would be useful like many other online platforms have.
There is an extension for sphinx https://sphinx-copybutton.readthedocs.io/en/latest/ but I don't know what steps are required to make it work with ReadTheDocs

## Acceptance Criteria:
- [x] Determine a ReadTheDocs compatible method for adding a copy-button
- [x] Implement the feature

rtapella:
example: https://docs.maap-project.org/en/code_copy_button/science/GEDI/GEDI_L2A.html

PR #265, opened 2023-07-18
add sphinx_copybutton

rtapella:
- add to requirements.txt so it installs
- add to extensions in conf.py
- fixes https://github.com/MAAP-Project/maap-documentation/issues/256


wildintellect (approved):
Do we need to test ReadTheDocs compatibility before merging?

PR #273, opened 2023-08-01
science_examples update

smk0033:
Noticed that "Nisar" needed to be updated to "NISAR" so it'll show up in the docs.


jjfrench (approved):
🚀

PR #300, opened 2023-08-10
Merge 3.1.0 Release Notes to main

rtapella:
- various science tutorial updates
- new page: Release Notes


smk0033 (approved):

rtapella:
I merged this but do I also need to tag it or something like that? Because it's released to `master`?

wildintellect:
@rtapella If I recall, tagging will make it also be the `stable` copy not just `latest`, so usually you would take `master` before merging new stuff in. If we missed it this time, not a big deal (and we could go tag the commit before the PR).

PR #319, opened 2023-09-05
collections.ipynb author edit

smk0033:
Added Aimee as an author for the already-updated searching collections tutorial https://docs.maap-project.org/en/latest/technical_tutorials/search/collections.html


smk0033:
😅😅

jjfrench (approved):

PR #328, opened 2023-09-07
Remove system_packages option from yml

rtapella:
Project will not build with the system_packages option in the .readthedocs.yml


rtapella:
doing a test build on readthedocs: https://docs.maap-project.org/en/remove-deprecated-yml-option/index.html

wildintellect (approved):

PR #331, opened 2023-09-18
chore: Integrate NISAR STAC endpoints

jjfrench:


smk0033 (approved):
Looks good! Only pointing this out because I'm not sure if this was intentional or not, but you have a variable as "item_assset" instead of "item_asset" (if that was intentional then ignore this!)

PR #370, opened 2024-01-09
Add workspace-upgrade instructions

rtapella:


docs/source/system_reference_guide/create_workspace.ipynb
anilnatha, line 1: Related to the statement:

>you are not sure which version of workspace to use, check the options by browsing the [Container Registry](https://repo.maap-project.org/root/maap-workspaces/container_registry).

It may be helpful to users if we could add a summary/description in the container registry for each available version.

sujen1412 (approved):

docs/source/system_reference_guide/create_workspace.ipynb
rtapella: in gitlab?

anilnatha: Yup, it's the only place I could find where our workspace versions are documented. @rtapella

PR #371, opened 2024-01-09
Adding memory profiling docs

omshinde:
Documentation for performing memory profiling any Python script.

### DIFF
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
@@ -5,7 +5,6 @@ from memory_profiler import profile
 """
 
 @profile
-
 def my_function():
     # Include each line of the script which needs to be profiled
     # under this function
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
@@ -1,7 +1,4 @@
 from memory_profiler import profile
-"""
-... Code here
-"""
 
 @profile
 def my_function():
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
@@ -1,0 +1,17 @@
+from memory_profiler import profile
+
+"""
+... Code here
+"""
+
+@profile
+
+def code_to_profile():
+    # Include each line of the script which needs to be profiled
+    # under this function
+    print("Hello, world!")
+    
+    return 0
+
+if __name__ == "__main__":
+    code_to_profile()
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
@@ -1,5 +1,4 @@
 from memory_profiler import profile
-
 """
 ... Code here
 """
@@ -8,9 +7,21 @@ from memory_profiler import profile
 def my_function():
     # Include each line of the script which needs to be profiled
     # under this function
-    print("Hello, world!")
     
     return 0
 
-if __name__ == "__main__":
+@profile
+def my_other_function():
+    # Include each line of the script which needs to be profiled
+    # under this function
+    
+    return 0
+
+def main():
     my_function()
+    my_other_function()
+    
+    #...
+
+if __name__ == "__main__":
+    main()
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling.py
@@ -6,7 +6,7 @@ from memory_profiler import profile
 
 @profile
 
-def code_to_profile():
+def my_function():
     # Include each line of the script which needs to be profiled
     # under this function
     print("Hello, world!")
@@ -14,4 +14,4 @@ def code_to_profile():
     return 0
 
 if __name__ == "__main__":
-    code_to_profile()
+    my_function()
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
@@ -8,13 +8,13 @@ fp=open('memory_profiler.log','w+')
 
 @profile(stream=fp)
 
-def code_to_profile():
+def my_function():
     # Include each line of the script which needs to be profiled
     # under this function
     
     return 0
 
 if __name__ == "__main__":
-    code_to_profile()
+    my_function()
     
     
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
@@ -1,0 +1,20 @@
+from memory_profiler import profile
+
+"""
+... Code here
+"""
+
+fp=open('memory_profiler.log','w+')
+
+@profile(stream=fp)
+
+def code_to_profile():
+    # Include each line of the script which needs to be profiled
+    # under this function
+    
+    return 0
+
+if __name__ == "__main__":
+    code_to_profile()
+    
+    
--- a/docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
+++ b/docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
@@ -1,11 +1,8 @@
 from memory_profiler import profile
-
 """
 ... Code here
 """
-
 fp=open('memory_profiler.log','w+')
-
 @profile(stream=fp)
 def my_function():
     # Include each line of the script which needs to be profiled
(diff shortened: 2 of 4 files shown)


wildintellect:
This is a good start. I think users might be a little confused about implementing in the ADE vs DPS.
- Can you clarify how to specify where the output file from the profiling ends up? Maybe also include a note about getting the output from a DPS job.
- [ ] Nice to have: a paragraph at the top about when to implement this, generally after you already have working code and you want to know what size worker to use in DPS, or to look for ways to reduce the RAM.

docs/source/technical_tutorials/user_data/memory-profiling-python.ipynb
```diff
@@ -0,0 +1,210 @@
+{
```
jjfrench, line 1: <p>Don't forget to give yourself credit for creating this documentation :) (format similar to other technical guides)</p><p> </p><p>Capitalization in the title and subtitles</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/371/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jjfrench, line 1: <p>Can we explain this a little more? Looks like we're adding a decorator to a new function. Can we add the decorator to an existing function or are users going to take all their lines of code and paste it under this function?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/371/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

smk0033: <p>Agreed! "Author(s):", "Date:", and "Description:" should be listed underneath the title 🙂 maybe to be consistent with other notebooks, change "References:" to a header saying "Additional Resources"</p>

smk0033, line 1: <p>My suggestion would be to shorten these headers some. Maybe something like "Step 1: Installation" and then in normal text below, put "Install <code>memory-profiling</code> in your existing working environment" and any additional details as necessary (I see Jamison's comment below on more elaboration for Step 3). I'd probably update that format for steps 1-4, I think the formatting for 5 is okay besides also updating the capitalization. Overall, I think it looks good though!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/371/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

omshinde: <p>Thanks, this is a good point. I've added explanation in the updated notebook.</p>

omshinde: <p>Thanks, @smk0033. I've added description and shortened the headings. </p>

omshinde:
Thanks @wildintellect @jjfrench @smk0033 for the review comments.

> Can you clarify how to specify where the output file from the profiling ends up? Maybe also include a note about getting the output from a DPS job.
@wildintellect I've specifically mentioned that this tutorial is for memory profiling in ADE and can be used for estimating DPS worker size.

> How about how to call an existing python file? possibly doing the calls from a notebook, and where to see the results.
Good point. Added a logging example for saving the output as a log file. It would be good for the users to save the memory profiling output and share later.

docs/source/technical_tutorials/user_data/memory-profiling-python.ipynb
jjfrench: <p>I'm still a little confused as to why we have to create a new code_to_profile function. Can we generalize the name of the function and describe that we can add the decorate to our already existing functions? Users might take it that they need to name their function code_to_profile and strip out all of their code from their existing functions</p>

omshinde: <p>Thanks. I agree as to not give any benefit of doubt for confusion. Changes:</p><ul><li>Removed Step 2, no need to add new function</li><li>Generalized function name to my_func()</li><li>Described decorator</li></ul>

docs/source/technical_tutorials/user_data/demo_memory_profiling.py
```diff
@@ -0,0 +1,17 @@
+from memory_profiler import profile
+
+"""
+... Code here
+"""
+
+@profile
+
```
chuckwondo, line 8: trivial: remove blank line after decorator -- Python convention is to have no blank lines between a decorator and the function it decorates

docs/source/technical_tutorials/user_data/demo_memory_profiling_logging.py
```diff
@@ -0,0 +1,20 @@
+from memory_profiler import profile
+
+"""
+... Code here
+"""
+
+fp=open('memory_profiler.log','w+')
+
+@profile(stream=fp)
+
```
chuckwondo, line 10: trivial: remove blank line after decorator

docs/source/technical_tutorials/user_data/memory-profiling-python.ipynb
chuckwondo, line 1: github won't allow me to comment on specific lines in this file, so I'm placing the comment on the file as a whole.

Overall, nice content! I have only some minor/trivial grammatical/syntactical comments:

1. as commented elsewhere, per Python convention, please remove blank lines after decorators
2. change "have a working code" to "have working code"
3. change "use `memory profiler` [tool]" to "use [memory-profiler]"
4. change "named as `my_function()`" to "named `my_function`" (i.e., drop "as" and also the parentheses)
5. change "with `@profile` decorator above the `my_function()`" to "with the `@profile` decorator above `my_function`"
6. change "allows to run" to "allows you to run"
7. change "By default the" to "By default, the"
8. change "as system `std_out`" to "as standard output"

omshinde: Thanks @chuckwondo for the detailed review. I've updated the notebook to accommodate your comments.

docs/source/technical_tutorials/user_data/memory-profiling-python.ipynb
chuckwondo: <p>To Jamison's point, perhaps we can add a bit more clarity here.</p><p> </p><p>How about something like the following for the code example, along with the sentence that follows?</p><p> </p><pre class="ql-syntax" spellcheck="false">@profile
def my_function():
    # Perform some potentially memory-intensive computation
    # ...
</pre><p> </p><p>You may add the <code>@profile</code> decorator to individual functions that you want to profile. This allows you to limit which parts of your program are profiled, thus limiting the volume of profiling output.</p><p> </p><p>However, this requires you to modify your code. If you wish to avoid modifying your code, particularly when it is not yet obvious which parts of your code may be consuming too much memory, simply add the <code>-m memory_profiler</code> to the <code>python</code> command.  </p>

docs/source/technical_tutorials/user_data/memory-profiling-python.ipynb
```diff
@@ -0,0 +1,278 @@
...
```
chuckwondo, line 1: <p>"Defining function call" is a bit of an awkward phrase. This also makes it seem like you're calling and profiling only 1 function.</p><p> </p><p>Perhaps something along these lines:</p><p> </p><pre class="ql-syntax" spellcheck="false">@profile
def my_function():
    # ...

@profile
def my_other_function():
    # ...

def main():
    my_function()
    my_other_function()
    # ...

if __name__ == "__main__":
    main()
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/371/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

chuckwondo: <p>Ugh! reviewnb totally ruins code blocks!</p>

omshinde: Thanks, @chuckwondo ..Noted with thanks and updated in the recent commit. Requesting re-review

chuckwondo:
Looking good. One more suggestion is to remove all occurrences in the python files and notebook of this:
```
"""
... Code here
"""
```

Given that the block is a comment block, it's confusing to see `...Code here` inside the comment. I think we can leave it out and readers will realize that there can be other code since we're showing only dummy functions.

omshinde:
Thanks @chuckwondo ..Updated the notebook accordingly.

chuckwondo (approved):
Looks good. Please make sure you get approval from 1 other reviewer as well, in case someone else has any further suggestions.

smk0033 (approved):
Looks good to me from the formatting perspective!

jjfrench (approved):

omshinde:
Close https://github.com/NASA-IMPACT/active-maap-sprint/issues/861 #371

PR #397, opened 2024-04-09
Fix blurry image

grallewellyn:
Renamed image to import_granules_1


rtapella (commented):
hm the import granules image change-diff says:
"File renamed without changes"

rtapella:
also it doesn't seem to blurry to me..

grallewellyn:
Strange, it looks fine here too: https://docs.maap-project.org/en/latest/system_reference_guide/faq/import_granules.html
Maybe we just close this PR then?

rtapella:
sounds good to me... maybe I added the better image in with develop separately and now it looks like it wasn't updated. I have a vague memory of just copying your updated image in to a different branch...

PR #404, opened 2024-06-12
Jobs output

grallewellyn:
Updated screenshots to reflect the jobs' output folders and also reflect the updated jobs UI


rtapella (approved):
Looks good

For the annotated image it might be easier to read if the notes are left-aligned but it's fine like this

grallewellyn:
@rtapella I agree, just made that change!

PR #419, opened 2024-08-01
added listJobs maap-py docs

marjo-luc:


docs/source/system_reference_guide/jobs_maappy.ipynb
```diff
@@ -135,6 +135,61 @@
...
+    "```\n",
+    "\n",
+    "Users may provide optional query parameters to filter the list on desired fields. The following example will return a list of jobs submitted by user `mlucas` \n",
+    "that all ran the `job-dps_tutorial_mlucas:main` algorithm:\n",
+    "\n",
+    "```\n",
+    "r = maap.listJobs(\"mlucas\", algo_id=\"job-dps_tutorial_mlucas\", version=\"main\")\n",
+    "```\n",
+    "\n",
+    "Users may also request just the list of jobs id's and their corresponding tags rather than full job details. This is controlled using the `get_job_details` option. By default `get_job_details` is set to `True`, but if set to `False` will return a compact list, as shown in this example:\n",
```
rtapella, line 166: how did we decide that get_job_details=False only returns JobID and Tag? Why not, for example, status?

docs/source/system_reference_guide/jobs_maappy.ipynb
```diff
@@ -135,6 +135,61 @@
...
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7100c385",
+   "metadata": {},
+   "source": [
+    "## List Jobs\n",
+    "\n",
+    "1. Use the `listJobs` method, specifying a username, to retrieve a list of jobs that user has submitted.\n",
```
rtapella, line 146: is there a listJobs() that gets everything? why not have listJobs(username=“mlucas”) as syntax?

docs/source/system_reference_guide/jobs_maappy.ipynb
```diff
@@ -135,6 +135,61 @@
...
+    "```\n",
+    "\n",
+    "These are the available query parameters:\n",
+    "\n",
+    "| Parameter       | Description | Type                        | Optional/Required? | Acceptable Values\n",
+    "|-----------------|-------------|-----------------------------|-------------------|----------|\n",
+    "| algo_id         | algorithm id e.g. `job-dps_tutorial_mlucas` | string  | optional          | Valid string |\n",
+    "| end_time        | job end time   | string e.g. `2024-01-01T00:00:00.000Z`  | optional          | Valid time string |\n",
+    "| get_job_details | Flag that determines whether to return full job details or just a list of job id's with their corresponding tags. Default is `True`.| bool  | optional  | `False` or `True`|\n",
+    "| offset         | pagination offset | number  | optional          | Integer |\n",
```
rtapella, line 184: is offset the page-number offset, or the job-number offset? (e.g. if there are 24 jobs with page_size of 20, is page_size only 1 or 2? or 1,2,3…24)?)

rtapella (commented):
Left some comments (by the way: I don't really have context for the changes or requirements for listJobs)

docs/source/system_reference_guide/jobs_maappy.ipynb
marjo-luc: We didn't really decide what gets returned in the abbreviated response. That's what the Mozart API returns. We could append arbitrary fields from the MAAP API, but we would have to request the full job details and parse the desired fields and add that to the MAAP API response.

docs/source/system_reference_guide/jobs_maappy.ipynb
marjo-luc: Offset is the job number offset from the job listing retrieved from Mozart/elasticsearch. So for example, if I make a listJobs request with offset=0 and page_size=2 and get these two job id's:

 0.   03695a4f-86f2-4418-9db0-f55e16a89901
  1.   f3a21979-18fe-4687-88d6-33f18cb5ad05

then I make another request but with offset=1 and page_size=1, I should get this:

f3a21979-18fe-4687-88d6-33f18cb5ad05

docs/source/system_reference_guide/jobs_maappy.ipynb
marjo-luc: `listJobs(username="mlucas")` is valid. I did not want to remove `listJobs("mlucas")` because users are currently using the latter. If we want to clean maap-py up (which I believe there are plans to) then we can revisit this and make sure to add deprecation warnings for any of the interfaces we change.

rtapella (approved):

PR #432, opened 2024-10-01
note about input management in run.sh

rtapella:


grallewellyn (approved):
Using the word "set" for files might be a little confusing, we could change to "multiple files" for clarity

grallewellyn:
^ This was just being nitpicky, it doesn't really matter @rtapella

PR #433, opened 2024-10-01
DPS tutorial updates 4.1.0

rtapella:


rtapella:
hm let me fix the conflicts

grallewellyn:
Why do we want to remove the section specifically for persistent storage? I know we repeat that information in "MAAP Storage Options" but wouldn't it be convenient to also have that information in FAQ?

rtapella:
I think it was out of date so I figured it's easier to just maintain it all in one place.
It seems more like a section in the System Reference Guide that should be pointed at by other documents, rather than in the GS guide or a FAQs.

grallewellyn:
Yes, I agree it is getting too chaotic to repeat information in multiple places

grallewellyn (approved):
Looks good to me

PR #441, opened 2024-10-15
fix: update conf.py #417

wildintellect:
This should resolve the RTD update compatibility. cc: @zacharyDez


rtapella:
https://about.readthedocs.com/blog/2024/07/addons-by-default/

rtapella (approved):
this change does not break anything.

Issue #442, opened 2024-10-15
Tags out of date


wildintellect:
We have not kept up with Tagging the docs when doing major merges to `Develop`

Each time we merge to `Develop` we're supposed to tag the current version so that it can be found in the history, and navigated to on ReadTheDocs. The current `Develop` should probably be tagged 4.1.0?

Additionally a new feature on ReadTheDocs might confuse users. 'Latest Version Notification' https://about.readthedocs.com/blog/2024/04/enable-beta-addons/ shows a popup when you go to `latest` (which is the Develop branch). However for us this is actually the most up to date documentation, there are no not yet implemented features in the docs.

@rtapella we might need to rethink the naming strategy of the branches and the configuration on the ReadTheDocs to make sure people are going to the best version of the docs and knowing it matches what they are using.

wildintellect:
@zacharyDez maybe you have some thoughts on how to straighten out our tags, and reconfigure RTD to not be confusing?

zacharyDez:
@wildintellect on the latest version notification, I did some research. I found [this discussion](https://github.com/readthedocs/readthedocs.org/issues/11474) on addons. It seems there isn't a clear option for disabling addons at the time. We can [change the configuration](https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html) used for themes, but that does not give us access to disable the warning. The simplest apporoach seems to be setting the `display` value to none on the `.rtd-version-warning` class:
`_static/custom.css`
```css
[...]
.rtd-version-warning {
  display: none;
}
```

rtapella:
When we update the docs for a release, we do make a documentation release as well (https://github.com/MAAP-Project/maap-documentation/releases). As a result, there is a tag in the potential list of docs for each release.

<img width="954" alt="Screenshot 2024-11-12 at 3 17 32 PM" src="https://github.com/user-attachments/assets/1d30ec3c-6faa-4d3e-8b17-4a21785c9035">

Potentially we should tag some additional merges to develop.

Latest for us is actually the master branch:
<img width="810" alt="Screenshot 2024-11-12 at 3 18 53 PM" src="https://github.com/user-attachments/assets/e433ea7e-3222-40e9-941b-536d659b4bf1">

rtapella:
(when we do a release, we merge develop -> master, cut a release (giving it a tag and version), and "latest" has the latest release of documentation)

rtapella:
Maybe latest should be develop, and stable should be master?

[...]

PR #444, opened 2024-10-15
Set display to none on rtd-version-warning tag to css

zacharyDez:
## What I Changed

Add display configuration for the version warning.

## How to Test It

Spin up the docs locally and verify that the warning message is no longer appearing.

## Other Notes

Relates to #442, but does not entirely close it as versioning is still an issue.

I found https://github.com/readthedocs/readthedocs.org/issues/11474 on addons. There seems to be no clear option for disabling addons at the time. We can [change the configuration](https://sphinx-rtd-theme.readthedocs.io/en/latest/configuring.html) used for themes, but that does not allow us to disable the warning.

### DIFF
--- a/docs/source/_static/custom.css
+++ b/docs/source/_static/custom.css
@@ -26,3 +26,7 @@
 .wy-nav-content-wrap {
   background: #fcfcfc;
 }
+
+.rtd-version-warning {
+  display: none;
+}


wildintellect:
It's configurable in the Admin Panel
Screenshot from 2024-10-15 13-40-12

But that's a temp fix, what I had in mine was cleaning up the tags so that RTD knows which branch is the "stable" which is what we currently call develop/latest

Issue #443, opened 2024-10-15
HLS science tutorial is blank


rtapella:
https://docs.maap-project.org/en/latest/science/HLS/HLSL30.html

- the page comes up blank
- a local build does not have the same issue

wildintellect:
We've narrowed it down. Known bug with the new Add-ons feature of ReadTheDocs and large files. This particular notebook being over 20 MB.

>"So far, two solutions are to output image files instead of inlining the images, or you might be able to disable Addons temporarily"

- https://stackoverflow.com/questions/79091590/read-the-docs-build-has-blank-html-for-some-pages
- https://matrix.to/#/!LVjngZMcDVmwNVkCqi:gitter.im/$KFeb_LmW1CW-UlJhy5EaiBlRL6jclHZ0SPbxRy4pIpc?via=gitter.im&via=matrix.org&via=cadair.com

We're looking into compressing images within the notebook. Looking at alternate methods to export the figures to individual files with separate compression and link them in via Markdown, not actual cell output which should shrink the file size.

PR #445, opened 2024-10-16
Reduce image size hls

rtapella:
https://docs.maap-project.org/en/reduce-image-size-hls/science/HLS/HLSL30.html

manually reduced the size of four images...


wildintellect (approved):
Looks Good. It's sad that we have to drop the interactive maps. Maybe we can find a way to put those back in later but the size savings is huge.

This should solve #443 

cc: @nmt28

rtapella:
Yes those maps are very large :/
If you clone the notebook into your Jupyter, you can still run the cell and get the map.

PR #453, opened 2024-10-26
Removing username maap-py listJobs & submitJob

grallewellyn:
Documentation for removing the need for adding username when submitting or listing jobs

- `maap._MAAP_HOST` feature seems to have been removed so removed the cell with this output

Now getting username from PGT token, not argument 

Making this change because of these PRs: https://github.com/MAAP-Project/maap-py/pull/108 https://github.com/MAAP-Project/dps-jupyter-extension/pull/23 https://github.com/MAAP-Project/jupyter-server-extension/pull/14


rtapella (commented):
looks good to me
should be merged into release/4.2.0 branch first, then develop when we release

PR #464, opened 2024-12-11
MAAP STAC Search in R

smk0033:
Github ticket: https://github.com/NASA-IMPACT/active-maap-sprint/issues/1040

Note: Tried to add @HarshiniGirish as an assignee since she wrote a majority of the notebook but it wouldn't let me


docs/source/technical_tutorials/search/maap_stac_r.ipynb
```diff
@@ -0,0 +1,734 @@
+{
```
hrodmn, line 1: <p>For the purposes of this notebook I don't think we need the <code>if(!is.null(collections$collections))</code> check since we know that this STAC API has collections. Maybe it would be better to show how to retrieve the metadata for a specific collection given an ID.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

hrodmn, line 1: <p>Same thing for this cell, I would remove the control flow that catches collections with no items. That might be helpful for the focused search section, but not here since these collections should all have items!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

hrodmn, line 1: <p>It would be helpful to mention that these are the hrefs for the COGs with the biomass estimates!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

hrodmn, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    http_urls = sapply(stac_query$features, function(x) {x$links[[5]]$href})
</pre><p>I would either remove the <code>http_urls</code> or section or make a note that these links are for the STAC item metadata (links for <code>rel='self'</code> I think).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

rtapella: Often people will use the tutorials as copy-paste example code. So if we expect people to use this notebook that way, I’d suggest leaving in the control-flow/checks on the collections. It’s also implicitly a good lesson that you can have empty Collections and that they need to be handled.

If we don’t expect anyone to copy/paste the code from this notebook, then I agree we could leave that out.

chuckwondo: <p>Since my knowledge of R is minimal, I could very well be wrong, but I suspect that <code>!is.null(collections$collections)</code> will <em>always</em> be true because <code>collections$collections</code> will <em>always</em> be a list. Of course, the list <em>could</em> be empty, but it should never be <code>NULL</code>. I suspect what you might want here is <code>length(collections$collections) &gt; 0</code> instead. It appears to me that there are a number of places where a null check is being performed, but a length check should be performed instead.</p><p> </p><p>Again, I could be totally off base, so someone please correct me if I'm wrong here.</p>

wildintellect: You might be right @chuckwondo 
We need to test against an endpoint with 0 collections, perhaps it will be easier to find a 0 items collection to test this on.

smk0033: <p>Oops, totally didn't realize those were the metadata files - I removed them!</p>

smk0033: <p>Tested with a collection that had 0 items and y'all were right - we updated to do link checks!</p>

docs/source/technical_tutorials/search/maap_stac_r.ipynb
```diff
@@ -0,0 +1,676 @@
...
```
chuckwondo, line 1: <p>I suggest eliminating use of <code>tryCatch</code> here because it is inconsistent with how your handling search requests everywhere else in this notebook.</p><p> </p><p>Further, you have 2 checks on the length of <code>items$features</code>, so perhaps the second block is unnecessary?   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

chuckwondo, line 1: <p>The following will cause an error if <code>length(items$features) == 0</code>:</p><p> </p>

<pre class="ql-syntax" spellcheck="false">first_item &lt;- items$features[[1]]
</pre><p> </p><p>Therefore, your previous length check was for naught.</p><p> </p><p>If you're going to include expressions that will fail on empty lists, then I suggest that you either <em>always</em> wrap such expressions within a length check, or <em>never</em> do so. In other words, don't mix the 2 approaches. Prefer consistency instead.</p><p> </p><p>In the context of this notebook, for simplicity and clarity, I would agree with Henry's suggestion to simply drop the length checks (previously null checks).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

chuckwondo, line 1: <p>Why are you using the <code>rstac::</code> prefix in this cell when you're not using it an any of the previous cells? Again, be consistent. Either <em>always</em> use the prefix or <em>never</em> use the prefix.</p><p> </p><p>Further, you defined <code>stac_endpoint</code> at the top of the notebook and use it for earlier queries, so use it here as well.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/464/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

smk0033: <p>Thanks, Chuck! Updated the code to account for the error. We went ahead and removed trycatch, but kept the length checks - the only reason being is because there is a collection with 0 items (AFRISAR_DLR2). However, if y'all prefer, we can still remove the length checks except for this updated cell (in case someone does use the empty collection I guess?)</p>

smk0033:
@chuckwondo @hrodmn since we had a long break, just wanted to tag y'all again for another review whenever y'all have some time

chuckwondo (approved):
LGTM!

smk0033:
Great, thanks! Forgot to update the rst files - will update the rst to add this notebook to the docs and then merge

PR #468, opened 2024-12-18
Docs/find r ref

zacdezgeo:
## What I Changed

Added reference in working with R in the pages listed below in how to test it.

## How to test it

Spin up docs locally and visit:
- http://[::]:8000/technical_tutorials/working_with_r/find_data_in_r.html
- http://[::]:8000/technical_tutorials/working_with_r.html
- http://[::]:8000/technical_tutorials/accessing.html
- http://[::]:8000/technical_tutorials/searching.html

## Other notes

Closes https://github.com/NASA-IMPACT/active-maap-sprint/issues/1044

### DIFF
--- a/docs/source/technical_tutorials/working_with_r/find_data_in_r.rst
+++ b/docs/source/technical_tutorials/working_with_r/find_data_in_r.rst
@@ -1,4 +1,4 @@
-Finding Data in R (OpenScapes)
+Finding and Accessing Data in R (OpenScapes)
 =====================================
 
 Visit the OpenScapes documentation here: `Finding Data in R (OpenScapes) <https://nasa-openscapes.github.io/earthdata-cloud-cookbook/how-tos/find-data/find-r.html>`_.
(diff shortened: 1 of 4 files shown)


wildintellect:
Is this the first time we've cross listed a tutorial under 3 different headings. Is that weird? Should there be a different way to send you to the R section?

zacdezgeo:
It's pretty clear where to go to work with R, it's under 'working with R'. The question seems more should the other sections of the guide be solely Python based? The find-R docs from OpenScapes have content on searching and accessing, so I linked them in those sections also. We may want to remove them.

docs/source/technical_tutorials/working_with_r/find_data_in_r.rst
wildintellect, line 1: this is good, we'll also link in the rstac with MAAP tutorial when @smk0033 finishes that

wildintellect (approved):
lets try it like this, if we don't like the way it comes out we can remove the extra links easily.

PR #491, opened 2025-04-23
titiler-pgstac in R

HarshiniGirish:


HarshiniGirish:
@wildintellect @smk0033 @hrodmn @zacdezgeo can someone please review this notebook when you find time. I want to make the changes and close the ticket before I leave for my break. Thank you

hrodmn:
@HarshiniGirish this looks really good! My only request for changes is to update the last section to use `ak_tilejson$tiles` instead of `gsub`ing the WMTS url to get the XYZ tile url.

wildintellect:
:thinking: do we want to make sure we can point people to WMTS urls if they wanted to use the tiles in QGIS? I think that would go in a different doc. I agree stick to simpler XYZ for Leaflet maps.

wildintellect:
@hrodmn I assigned final review to you.

HarshiniGirish:
Thankyou @hrodmn! up for another review.

hrodmn:
Thanks for making those changes @HarshiniGirish! I found a few more things that need to be addressed before we merge. You will also need to add an entry for the new notebook to `docs/source/technical_tutorials/working_with_r.rst` to make sure this notebook gets included when we rebuild the docs website.

It would be really helpful if you could make sure the notebook can be cleanly rendered from top to bottom (run all cells) after restarting the kernel - that is what is going to happen when we merge the PR and rebuild the docs site so we need to make sure that each notebook can be rendered cleanly without any errors!

HarshiniGirish:
Thank you @hrodmn. Please let me know if anything else has to be addressed!

HarshiniGirish:
thank you @hrodmn made the change. Not sure it is what you were looking for. I apologize for the disorder here

HarshiniGirish:
Thank you @hrodmn! made the updates!

hrodmn (approved):
Looks great - thanks @HarshiniGirish

PR #497, opened 2025-05-15
fix job-offline doc

rtapella:
- split up failed and offline
- describe what offline means


grallewellyn (approved):
Thanks for doing this, this is much more clear

rtapella:
also added the copc_access notebook to the "access" table of contents so that it shows up in the navigation

PR #501, opened 2025-06-03
Convert STAC Doc to rst

smk0033:
Converted STAC Metadata Recs document from markdown to rst. Docs site currently does not support markdown


docs/source/technical_tutorials/user_data/stac_metadata.rst
```diff
@@ -0,0 +1,174 @@
+MAAP STAC Metadata Recommendations
+==================================
+
+MAAP uses the STAC (SpatioTemporal Asset Catalog) specification, which
+provides a common structure for describing and cataloging geospatial
+data `(STAC) <https://stacspec.org/en>`__.
+
+At a minimum, MAAP metadata should conform to STAC specification
```
wildintellect, line 8: This document is a little confusing, because at the title says recommendations, but then later mostly lists requirements.

I think we should be clear:
- Mark which things are required by STAC - and tools to validate
- Mark which things are "Preferred" if you want to put something in MAAP STAC
- Mark the properties that have a match in CMR UMM (Bonus to link)

Maybe we use little icons or table columns.

docs/source/technical_tutorials/user_data/stac_metadata.rst
```diff
@@ -0,0 +1,174 @@
...
+
+   -  ``type``: **Required**. The media type of the asset.
+
+      -  `Common Media Types in
+         STAC <https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#common-media-types-in-stac>`__
+
+-  ``stac_version``: **Required**. The STAC version the collection
+   implements.
+
+STAC Item-Level Metadata Fields
```
wildintellect, line 134: Link to https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md

docs/source/technical_tutorials/user_data/stac_metadata.rst
hrodmn: I think it would be fine to declare non-required STAC fields to be required for the MAAP STAC if we want to increase the quality of our STAC metadata! 

This is our chance to steer users towards a more rigorous specification - listing the optional but preferred fields as such will probably lead to individuals choosing to skip them. We can always choose to let something through even if it doesn't meet these guidelines, but I am in favor of listing all of these fields as required to nudge users in that direction.

We should just distinguish the fields that are truly optional due to availability of information. Like `sci:doi` which might not exist yet so you can't include it.

docs/source/technical_tutorials/user_data/stac_metadata.rst
```diff
@@ -0,0 +1,174 @@
...
+
+STAC Collection-Level Metadata Fields
+-------------------------------------
+
+-  ``id``: **Required**. The unique identifier of the collection.
+-  ``version``: **Required**. The version of the data. This is different
+   than the ``stac_version`` element.
+-  ``title``: **Required**. A short descriptive title of the datatset.
+   This should differ from the ``id``.
+-  ``sci:doi``: **Required if available**. The DOI of the collection
```
hrodmn, line 34: We should ask users to include other fields from the scientific extension if `sci:doi` is not available. Using `sci:citation` like we did for icesat2-boreal where the publication was still in prep.

hrodmn (commented):
As I say in the thread below, I am in favor of pushing all of these fields as required (with a few exceptions).

I think it would be useful to include the [icesat2-boreal-v2.1 metadata](https://github.com/MAAP-Project/icesat2-boreal-stac/blob/main/examples/agb/collection.json) as an example (printed inline) since we went through that one very carefully and implemented the UMM-compliant fields. Let's hold that one up as the standard here - printing the JSON inline and linking to that repository should give users a good starting point for a new dataset.

docs/source/technical_tutorials/user_data/stac_metadata.rst
wildintellect: @hrodmn Yes I just wanted to be clear what is STAC required (that a Validator might catch), and what is MAAP asking for in addition to that to ensure good quality metadata.

docs/source/technical_tutorials/user_data/stac_metadata.rst
smk0033: Sure! I'm going through and adding superscript to each element

```
- ``element``:sup:`1`: required by STAC

- ``element``:sup:`2`: required by UMM

- ``element``:sup:`3`: not required by STAC or UMM, but still required by MAAP
```

wildintellect: seem ok for now - at least to me it's much clearer.

smk0033:
I also still need to look into UMM-Var😅

wildintellect:
We decided in a meeting to address CEOS differences in the MAAP Data Requirements Document 1st, then port those changes to a future update of this page.

> * CEOS has the full hierarchies listed for science keywords - is this something we want to recommend or implement?
Not unless CMR/UMM also requires this. We'd also have to look into how implement in STAC in a sane way that doesn't overly duplicate.

>     * CEOS recommends absolute href paths. I think we already do this, so I added a general note at the top
That makes sense, pretty sure we need absolute paths to find things in Buckets. There's a minor technical question as to where the paths get made that way. Example: in DPS output, the paths are relative inside the job, and the STAC ingestion tool @hrodmn made will convert them to absolute.

>     * CEOS recommends having created, updated, and published datetime properties at the collection level. I think UMM requires provider dates the granule level, do we also want to add for the collection? It also seems like the ICESat item has a "created_datetime" but not updated or published unless I've overlooking it

how does published differ from created?
Updated - we could do that to track if edits happen to the metadata, probably something our tools would autopopulate.

smk0033:
Also the created/published is a good question. I'm assuming it'd probably be something along the lines of a STAC metadata record being created (maybe in a dev env), and then the publish date would be when the data/metadata is made available in prod?

hrodmn (approved):

PR #535, opened 2025-12-08
external_access_from_hub

HarshiniGirish:


wildintellect (changes requested):
I don't think we are merging anything from `maap-hub-docs` into the main docs yet. I'll still review this notebook but just for additional fixes to happen in it's current location.

Where in the Nav should this page be list, I don't see it on https://docs.maap-project.org/en/hub/technical_tutorials/accessing.html

docs/source/technical_tutorials/access/external_access_from_hub.ipynb
```diff
@@ -0,0 +1,1310 @@
+{
```
wildintellect, line 1: <pre class="ql-syntax" spellcheck="false">Line #5.    matches = fs.glob(pattern)
</pre><p>Please don't scan the bucket for files, just hard code an example like the example above, or does the filename keep changing? Reason, we want users to rely on STAC and CMR to find files not bucket listing which won't always work depending on the DAAC and Collection.</p><p> </p><p>It also adds a lot of complication to the code example.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/535/discussion/#comment-2600614982'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='13'/>

wildintellect, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    
</pre><p>Add a comment about always optimizing the cache, and link to the Cloud Native guide on this topic.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/535/discussion/#comment-2600614984'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='15'/>

wildintellect:
Seems we need a companion doc or link to VEDA docs showing how to use EDL to connect to other DAACs but not use a netrc file which isn't particularly secure.

HarshiniGirish:
I’m going to close this ticket once merged, I’ll write a separate notebook that shows how to sign in with Earthdata Login, and then I’ll add a link to it from the current examples. @wildintellect

wildintellect:
@HarshiniGirish https://github.com/MAAP-Project/maap-documentation/pull/537 now shows how to login, so you don't need to do it.

So I think this one is done now.

PR #536, opened 2025-12-08
Initial commit of 2i2c notes

grallewellyn:
Can users have 2 workspaces running? For me, when I press add server it just takes me back to the server I already have running. If I have nothing running and press add new server, it creates a server without a name cc @wildintellect 

Note some of links won't work until I get approval to merge: https://github.com/MAAP-Project/maap-workspaces/pull/183

### DIFF
--- a/docs/source/system_reference.rst
+++ b/docs/source/system_reference.rst
@@ -9,6 +9,7 @@ It is often referenced from the application-focused Science and Technical tutori
   :maxdepth: 2
   :caption: System Reference Guide:
   
+  system_reference_guide/2i2c_hub.ipynb
   system_reference_guide/create_workspace.ipynb
   system_reference_guide/custom-environments.ipynb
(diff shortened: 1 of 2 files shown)


grallewellyn:
Cancelling because I should checkout from the maap-hub-docs branch not develop: https://github.com/MAAP-Project/maap-documentation/pull/537

wildintellect:
> Can users have 2 workspaces running? For me, when I press add server it just takes me back to the server I already have running. If I have nothing running and press add new server, it creates a server without a name cc @wildintellect

Yes you can have more than 1 running, users have to know to go to `https://hub.maap-project.org/hub/home` to see the list. The first item, which is the default is never named, and if running will automatically come up.
You have to type in a name before clicking Add New Server

These are stock JupyterHub features, we should find Jupyter or 2i2c docs that explain.

PR #534, opened 2025-11-25
feat: outline new guide to disk types and usage

wildintellect:
1st draft of disk usage guidance.

PR #537, opened 2025-12-08
2i2c overview

grallewellyn:
Can users have 2 workspaces running? For me, when I press add server it just takes me back to the server I already have running. If I have nothing running and press add new server, it creates a server without a name cc @wildintellect

Note some of links won't work until I get approval to merge: https://github.com/MAAP-Project/maap-workspaces/pull/183


wildintellect:
See answer in https://github.com/MAAP-Project/maap-documentation/pull/536#issuecomment-3629076853

wildintellect:
<p>Content wise this seems fine, and we should remember to cross link it with the content in #534</p><p>We can revise wording and add more content as we get feedback from users.</p>

grallewellyn:
@wildintellect @smk0033 This is ready for review now, I changed mentions and images of eclipse che to 2i2c

docs/source/getting_started/getting_started.ipynb
```diff
@@ -29,22 +29,14 @@
    "source": [
```
wildintellect, line 29: <p>Highlight, circle or arrow to the EDL Login link?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/537/discussion/#comment-2624112995'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

wildintellect, line 29: <p>"If you are not what image you want," &lt;- not sure?</p><p> </p><p> </p><p>"You need at least 14.8GB RAM to use conda" -- this is incorrect. You need at least 4 GB of RAM to install additional packages. However the image itself will run on the 1.8GB (2GB) - maybe we need to borrow text from <a href="https://github.com/MAAP-Project/maap-documentation/blob/1ad471d15166ba9295d10e0cb6ea454671bcde7c/docs/source/system_reference_guide/faq/ade_to_hub.ipynb" target="_blank">https://github.com/MAAP-Project/maap-documentation/blob/1ad471d15166ba9295d10e0cb6ea454671bcde7c/docs/source/system_reference_guide/faq/ade_to_hub.ipynb</a> ... "Pick the smallest RAM/CPU allocation you need when booting a workspace. Smaller instance boot faster and cost less, so we can redirect that spend towards more compute" , I'm realizing we need to fix the default RAM selection to be 8 GB</p><p>2 GB is good for just opening files, editing text</p><p>4 GB is the minimum if you need to test installing Conda packages* -- if you frequently need additional packages contact the Support team to learn how to create a persistent Custom env. </p><p>8 GB+ depends on science needs, you can always restart with more RAM, Minimum recommended for R sessions.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/537/discussion/#comment-2624112996'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='4'/>

docs/source/system_reference_guide/custom-environments.ipynb
```diff
@@ -11,13 +11,10 @@
     "\n",
```
wildintellect, line 11: <p>14.8GB RAM  -&gt; 4 GB</p><p> </p><p>This page should actual direct people to </p><p><a href="https://docs.openveda.cloud/user-guide/scientific-computing/custom-environments.html" target="_blank">https://docs.openveda.cloud/user-guide/scientific-computing/custom-environments.html</a></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/537/discussion/#comment-2624113000'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='1'/>

wildintellect, line 11: <p>We need to note, these are Python only instructions now. We'll need a new doc to describe R customization.</p><p> </p><p>And once again link to <a href="https://docs.openveda.cloud/user-guide/scientific-computing/custom-environments.html" target="_blank">https://docs.openveda.cloud/user-guide/scientific-computing/custom-environments.html</a></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/537/discussion/#comment-2624113003'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

docs/source/getting_started/getting_started.ipynb
grallewellyn: Yes, I was trying to simplify this for non technical users even though it should have read "You need at least the 14.8GB RAM **option** to use conda **to install packages**" since I assumed users would mostly be using conda to install/ uninstall packages. 
I tried to create an 8GB RAM option but was suggested not to: https://github.com/2i2c-org/infrastructure/pull/7052#pullrequestreview-3398195848 which leaves the 7.4GB default option. I tested this option for a couple conda packages and it fails on `tensorflow` for one example and you previously said you need to use an 8gb or bigger instance for conda to work 

I would imagine it is very common for users to be using conda and I think our default option should allow them to safely use conda (so 14.8GB). The error when you don't have enough RAM to be installing with conda is very confusing because the command just exits, so I don't want users to be running into this issue

wildintellect: That's really interesting, I've always gotten it to work with 3.7 GB (~4 GB), though maybe I said 8GB which means the 7.4GB instance, but I never tried Tensorflow.
From what I've observed the crash actually happens at the package search and solving stage. Once it solves, the installs work. 

PS: I just round the RAM numbers up to the nearest when talking about it. I've been asking for years for us not too have such weird numbers listed (should probably round down to the nearest .5). I think that change is coming soon.

grallewellyn: Should we keep the suggestion at 14.8GB RAM option then so we can be sure that conda works? If adding an option in between 7.4GB and 14.8GB isn't an option? 
We could run a cost estimate on this if we think it will be a problem

wildintellect: Per slack discussion, we can leave it as is. https://grafana.maap.2i2c.cloud will show us how much users actually use, so we can adjust the default based on real usage. 

Noting that `conda` has gotten greedier in it's ram usage, possibly because the conda-forge archive has gotten bigger, it loads the whole archive in memory when solving, and the also seems to take even more ram when a solve fails.

docs/source/system_reference_guide/custom-environments.ipynb
wildintellect: ignore the ram part, but still direct people to the link.

grallewellyn: Made the change, this is ready for review now!

docs/source/system_reference_guide/custom-environments.ipynb
```diff
@@ -1,156 +1,13 @@
 {
```
wildintellect, line 1: <p>Hmm, the instructions include R already. repo2docker actually supports even more than that, we have a Julia example too <a href="https://github.com/US-GHG-Center/custom-images/tree/julia/retrievaltoolkit-docker" target="_blank">https://github.com/US-GHG-Center/custom-images/tree/julia/retrievaltoolkit-docker</a></p><p> </p><p>Agree more docs are coming, but Python and R are ready to use.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/537/discussion/#comment-2625152596'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='0'/>

docs/source/system_reference_guide/custom-environments.ipynb
wildintellect: oh I see the confusion, I was saying the instructions you wrote were python only, but the linked info is Python and R.

docs/source/system_reference_guide/custom-environments.ipynb
```diff
@@ -11,21 +11,18 @@
     "\n",
```
wildintellect, line 11: <p>I don't think we want people doing this, suggest removing entirely in favor of repo2docker method.</p><p> </p><blockquote>However, this stores this environment files in <code>/srv/conda</code>, which is a directory that is recreated when the workspace restarts, and so packages stored in that directory by the user are lost if a workspace restarts. Therefore, you want to specify another storage location in your user directory with the <code>--prefix</code> parameter</blockquote>

<pre class="ql-syntax" spellcheck="false">conda env create -f env-example.yml --prefix /home/jovyan/env/env-example 
</pre><blockquote>and to activate it :</blockquote>

<pre class="ql-syntax" spellcheck="false">conda activate /home/jovyan/env/env-example 
</pre><p> </p><p> </p><p>Now I'm wondering how we deal with needing the MAAP extensions and <code>maap-py</code> , or do we just say for now -- if you want to manage jobs use one of prebuilt images? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/537/discussion/#comment-2625200526'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='2'/>

wildintellect (approved):

PR #539, opened 2025-12-11
Delete Old NISAR Tutorial

smk0033:
Removing outdated NISAR tutorial completely from repo. It was previously just removed from the "science_examples.rst" file, but the tutorial does still come up when you search for NISAR in the docs site.


wildintellect:
:thinking: wondering if we need to also do this PR directly against `master` branch which is the alias for `latest`

smk0033:
It looks like if I open a new PR to compare this branch to master, it wants to include several other commits since the last push to master... maybe we should try and do another release next week with new updates, and try and include Hub docs then as well? @wildintellect

wildintellect:
I agree, that we could do a new push from `develop` to `master` since there are other useful updates. `Hub` should be skipped - it will remain a separate branch for a few months, until the Hub becomes the default interface -- at least that's the current plan since some edits related to the Hub would wipe out ADE documentation.

HarshiniGirish (approved):

smk0033:
Sounds good! I'll merge this and open a PR to merge develop into master

PR #540, opened 2025-12-12
v4.3.1 release

smk0033:
Some updates:
- new Earthcare notebook and updates
- remove outdated NISAR tutorial from repo so it can no longer be searched for in docs site
- new OPERA-DISP notebook
- DAAC access updates
- new GEE embedding notebook

Note: this release should not have any MAAP Hub documentation


smk0033:
So far I think this is looking pretty good, I don't think I see any new Hub documentation. I've noticed though that the leaftlet maps don't seem to be rendering in working_with_r/vector_data_visualization and working_with_r/visualizing_with_titiler-pgstac. I'm also not getting an image for the last cell in science/GEE/gee, but not sure if it's just me or not

wildintellect:
Looks like those maps were never working, so that should be a blocker on this.
- https://docs.maap-project.org/en/latest/technical_tutorials/working_with_r/visualizing_with_titiler-pgstac.html
- https://docs.maap-project.org/en/develop/technical_tutorials/working_with_r/visualizing_with_titiler-pgstac.html
cc: @HarshiniGirish @hrodmn

wildintellect:
@hrodmn @HarshiniGirish we can an open a new ticket to fix the maps, my suspicion is a difference in how the notebook records the executed R code, unclear if it stores rendered html+js the same way as python.

hrodmn:
I filed a ticket for the failed maps here https://github.com/NASA-IMPACT/active-maap-sprint/issues/1326

grallewellyn:
The earthcare notebook fails for me with `"['assets.enclosure_1.href', 'assets.enclosure_2.href'] not in index"` in the build assets table cell
The vector data visualization notebook assumes tmap is installed which it is not in the R image -> should this be in the R image? 
direct_access.ipynb says it was run in a vanilla workspace which is no longer an option, did you mean Python

wildintellect:
`tmap` is actually really hard to install in the current ADE image, I forget exactly, but you have to pin an old version. It will work in the hub.

Direct_access should probably actually reference Pangeo, that's the default python image we want people to use.

We'll look into the EarthCARE issue.

smk0033:
Hahah that'd explain why I was struggling so much getting tmap to work when testing the vector data visualization notebook with Henry's fix. I'll run it in the Hub instead

grallewellyn:
Yes, I meant adding to the Hub images, I just tested and added tmap to our maap-workspaces repo so it will be in our next release 
I pushed the image in case you would like to use it Sheyenne `mas.dit.maap-project.org/root/maap-workspaces/2i2c/r:add-tmap`

smk0033:
Ran the docs locally and updates seem to have been made and render correctly now. Should be ready for another review! As an FYI, I'm out until January 12th. Feel free to make any other updates or merge if everything looks good

grallewellyn:
I made a branch `small-changes-tutorials-release` with changes. Here are my notes from running:
**Earth care notebook**-> can run everything except I can’t generate a token at https://portal.maap.eo.esa.int/ini/services/auth/token/ (probably because I am not ESA)

**Gee**->  You need to create a new project to access https://code.earthengine.google.com/?scriptPath=users%2Fmmacander%2Fveg_structure%3Asentinel_1%2Fseasonal_s1_tile_export_v2 Is there another way to make this information more accessible to users? 

It is unclear what code that should go in sys.path.append('/projects/code/icesat2_boreal/lib') 

Adding credentials to s3://maap-ops-workspace/{USER}/credentials is not secure since other users can see the credentials. I thought we had a plan to not allow users to see other users buckets, but I tested and that has not been implemented yet. I stopped checking if the notebook runs after this point but checked the code 

**direct_access**-> Do users need to be logging into their own AWS account to get the temporary credentials? Because the role that we default assign users for a 2i2c session doesn’t have access to assume other roles. We also don’t have a maap-data-reader role on our AWS account either. I think we need more documentation on how to do this, I am not sure if there are more docs we can just link to 

**vector_data_visualization** → works for me 
**visualizing_with_titiler-pgstac** → works for me

wildintellect:
> I made a branch `small-changes-tutorials-release` with changes. Here are my notes from running: **Earth care notebook**-> can run everything except I can’t generate a token at https://portal.maap.eo.esa.int/ini/services/auth/token/ (probably because I am not ESA)
> 

We know the token retrieval is currently broken.

> **Gee**-> You need to create a new project to access https://code.earthengine.google.com/?scriptPath=users%2Fmmacander%2Fveg_structure%3Asentinel_1%2Fseasonal_s1_tile_export_v2 Is there another way to make this information more accessible to users?
> 
> It is unclear what code that should go in sys.path.append('/projects/code/icesat2_boreal/lib')
> 
It comes from https://github.com/lauraduncanson/icesat2_boreal/tree/master/lib but we should remove this dependency.

> Adding credentials to s3://maap-ops-workspace/{USER}/credentials is not secure since other users can see the credentials. I thought we had a plan to not allow users to see other users buckets, but I tested and that has not been implemented yet. I stopped checking if the notebook runs after this point but checked the code
> 

It's a temporary token, but looks like for notebooks it's good for 1 week. This method was originally developed before MAAP Secrets were available, but we also need to verify if Secrets can hold the entirety of the text. The DPS job would also need to be adjusted to read from secrets instead of S3.

@smk0033 We have 2 options:
1. Remove from the navigation - so the tutorial exists but isn't easy to find until we fix.
2. Remove the notebook too, making sure we have the latest in a branch where we can revise and resubmit once fixed.
Also noting that setting up a GEE account is not so simple anymore and this script assumes you already have one.
cc: @grallewellyn @omshinde

> **direct_access**-> Do users need to be logging into their own AWS account to get the temporary credentials? Because the role that we default assign users for a 2i2c session doesn’t have access to assume other roles. We also don’t have a maap-data-reader role on our AWS account either. I think we need more documentation on how to do this, I am not sure if there are more docs we can just link to
> 
This works differently on the Hub and is documented in the Hub branch.

omshinde:
I'm fine with both the options, preferably 1 and adding a big disclaimer at the top of the GEE notebook - "This notebook assumes that you have your GEE account setup. If not, refer here: [GEE docs]". We can share this notebook only with the users when they ask for it.

smk0033:
Updates seem to be reflected! Looks like some tickets will need to be made and more updates made to the docs later, but is there any additional feedback on this release?

wildintellect (approved):
@smk0033 I think you're right that this is ok to merge now, as long as we ticket those other todo items.

grallewellyn (approved):

PR #542, opened 2025-12-15
fix: load Proj and Leaflet js and css for R notebooks

hrodmn:
The R leaflet package embeds the javascript differently than the Python version so we need to manually load some dependencies to make the leaflet widgets work in R notebooks.

I tested it by building the docs site locally - the widgets work fine after loading these libraries (no change to the notebook code required).

Closes https://github.com/NASA-IMPACT/active-maap-sprint/issues/1326

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -16,6 +16,21 @@ import os
 
 def setup(app):
     app.add_css_file('custom.css')
+    # Add Leaflet CSS and JS for R htmlwidgets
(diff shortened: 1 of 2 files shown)


smk0033:
Confirming that I get output with visualizing_with_titiler-pgstac! Trying vector_data_visualization, but struggling with the tmap package for some reason

smk0033:
Still struggling some with tmap, but when I render this branch locally, it seems to work so I won't rerun that. Thanks, Henry!

smk0033 (approved):

PR #544, opened 2025-12-17
added repo2docker example MAAP images and MAAP extensions

grallewellyn:


wildintellect (approved):
Since the packages are Jupyter extensions, I'm wondering if we should work on conda-forge packaging? But that's for a new ticket/idea on the backend. The content here seems fine.

PR #546, opened 2026-01-05
small edits to tutorials release

grallewellyn:
Notable changes: 
- username is no longer needed for maap-py.submitJob
- I added the note about setting USER because I was confused at first if USER was the owner of the algorithm or the user running this notebook
- earthengine-api is not already in the pangeo image, should it be added?


wildintellect:
> Notable changes:
> 
>     * username is no longer needed for maap-py.submitJob
> 
Good to know.

>     * I added the note about setting USER because I was confused at first if USER was the owner of the algorithm or the user running this notebook
> 
It should be the person running the notebook. But that raises and interesting issue. I think hard coding the output directory to S3 as an algorithm variable is probably not a good idea. We probably need open a bug on the algorithm repo to fix that.

>     * earthengine-api is not already in the pangeo image, should it be added?
At this time no, I think we'll keep it as a user add-on, trying to balance default vs kitchen sink. Because of that maybe we need the `conda install` command in a code block at the top.

grallewellyn:
Added conda install command at the top 

> I think hard coding the output directory to S3 as an algorithm variable is probably not a good idea. We probably need open a bug on the algorithm repo to fix that.

How else would specifying the output directory be done?

wildintellect:
> How else would specifying the output directory be done?

:thinking: `/output` in a DPS job automatically goes to a users private bucket, and the DPS API getresults should provide the path. At least that's the traditional way. There is 1 DPS arg that does allow modifying the output path. I'll need to look more in depth at the algorithm before commenting more -- it might be doing direct S3 writes bypassing DPS file handling. But this is an algorithm change first. The notebook as written is correct given the current algorithm.

smk0033:
For the GEE notebook, maybe we could use the following code to get the username for environment setup? This came from another notebook

> maap = MAAP(maap_host="api.maap-project.org")
> USER = maap.profile.account_info()["username"]
> USER

grallewellyn:
@smk0033 I agree and I just made the change!

wildintellect:
@grallewellyn since you're modifying the GEE with this PR can you also delist it from the Navigation as noted in https://github.com/MAAP-Project/maap-documentation/pull/540#issuecomment-3750394870

grallewellyn:
Done, do we want to add this disclaimer: "This notebook assumes that you have your GEE account setup. If not, refer here: [GEE docs]" If so, what is the link to the GEE docs? https://courses.spatialthoughts.com/gee-sign-up.html?

smk0033:
I think the disclaimer would be great to have! I also found these docs: https://developers.google.com/earth-engine/guides/access
cc: @omshinde

smk0033 (approved):

PR #549, opened 2026-01-08
Update notebook to have ESA long-term token

omshinde:
Fix https://github.com/NASA-IMPACT/active-maap-sprint/issues/1333
- Updates the earlier ESA token access method to the new long-term (90 days) -> short-term (notebook/script execution) token access method. See [here](https://portal.maap.eo.esa.int/ini/kb/books/enhanced-data-access/page/using-the-long-lasting-token-to-access-and-download-data) for details.


omshinde:
Thanks @wildintellect. That's a valid point. I updated the ESA CCI notebook and the Earthcare notebook to reflect this. The ESA BIOMASS notebook will be changed in its respective PR.

wildintellect:
<p>make it clear where a user should save the file.</p>

---

 <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/549/discussion/#comment-3724975633'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

wildintellect:
Overall the fix is ok, I'm not sure the text is clear where and how to save the token to a text file in their home directory.
Later we should open a second ticket to enhance these code examples with more comments.

hrodmn (commented):
Since we are using same ESA token credentials file approach in multiple tutorials (this one and the Biomass notebook) I think we need to consolidate the documentation and create a single guide to managing ESA credentials then reference it as a prerequisite to the science examples.

You could make the new ESA token notebook in this PR but I think it would be fine to merge this PR as-is and the Biomass example PR then handle the consolidation in a follow-up PR after the two ESA data access PRs are merged to the develop branch.

We could do better than a long-lived credentials.txt file. In the longer term (PI 26.2?) I think we should:
- update the Hub/ADE images to define the `ESA_CLIENT_SECRET` and `ESA_CLIENT_ID` constants as environment variables
- add a `maap-py` function that can be used to retrieve the access token

wildintellect:
> We could do better than a long-lived credentials.txt file. In the longer term (PI 26.2?) I think we should:
> 
>     * update the Hub/ADE images to define the `ESA_CLIENT_SECRET` and `ESA_CLIENT_ID` constants as environment variables
> 
>     * add a `maap-py` function that can be used to retrieve the access token

You're right that we should move this discussion to future planning. The downside of leveraging maap-py in that way, is that the code becomes less universal ... which might be ok if a MAAP account is required for ESA access.

omshinde:
The NASA EDL authorization works for getting the ESA long-term data access token. The notebooks are ready to be reviewed with following additional changes to be done later-

* update the Hub/ADE images to define the `ESA_CLIENT_SECRET` and `ESA_CLIENT_ID` constants as environment variables

* add a `maap-py` function that can be used to retrieve the access token

wildintellect:
@omshinde both of those sound like they need to be tickets to be assigned to JPL.

HarshiniGirish (approved):

HarshiniGirish:
thanks @omshinde. they are good to go!

hrodmn (approved):
These changes make sense - I still think we should create a single notebook that shows how to get and store the token then reference it in the various science examples notebooks, but that can happen later.

PR #552, opened 2026-01-27
NISAR notebook

HarshiniGirish:


omshinde (commented):
Thanks, Harshini! Added a couple of minor comments and a major one for the visualization.

wildintellect:
I have a few questions:
1. Why are we downloading the file and using h5py directly to read the data, after we already setup Xarray data tree and fsspec to read directly from S3? Shouldn't we use that same pattern and opt for the cloud first method, even if it's not speed optimized yet?
2. Why is reprojection required for visualization, what's wrong with plotting in the native `epsg = 32615` for a single granule?

HarshiniGirish:
Thanks @wildintellect. I looked up reference from the OPERA notebook and made it more cloud optimized workflow. Its up for another review.

omshinde:
> Why is reprojection required for visualization, what's wrong with plotting in the native epsg = 32615 for a single granule?

Thanks @wildintellect .. There is no problem. We wanted to do it so that it can be visualized in WGS 84, also following the https://docs.maap-project.org/en/latest/science/NISAR/Simulated_NISAR.html example whic showed visualization after reprojection. No blocking preference there.

omshinde (approved):
Looks good to me. Thanks @HarshiniGirish

PR #560, opened 2026-02-06
Fix redundant names

omshinde:
- Fix the same names for ESA BIOMASS notebooks


wildintellect (approved):
Minor for next iteration - names are kind of long. (token secured) feels unnecessary in the title.

PR #562, opened 2026-02-12
Nisar dem notebook

HarshiniGirish:


nemo794:
Would it make more sense to use the NISAR DEM vrt files to access/subset the COGs? (vs. searching for individual COGs and manually combining/subsetting them.)

ASF just released documentation/tutorials for this: https://nisar-docs.asf.alaska.edu/nisar-dem/#vrt-reference-files

HarshiniGirish:
https://gist.github.com/HarshiniGirish/9d84ec88a0ecacd300c51adb5688b4a5 

When I try to open the HTTPS VRT https://nisar.asf.earthdata, cloud.nasa.gov/NISAR/DEM/v1.2/EPSG4326/EPSG4326.vrt, I get HTTP 302. Can you please take a look and confirm what’s happening? @nemo794 @wildintellect

wildintellect:
302 is a redirect, that wouldn't surprise me, but also I'm wondering if the EDL login issues today are also involved.
So we'll need to do a little more research on how to make the vrt work.

nemo794:
> 302 is a redirect, that wouldn't surprise me, but also I'm wondering if the EDL login issues today are also involved. So we'll need to do a little more research on how to make the vrt work.

Hmm, the VRTs should be stored in the exact same way (redirects and all) as the COGs which are currently used in the notebook.

If there were EDL issues today, probably best to simply wait until those are resolved to try again with the VRT?

HarshiniGirish:
Quick update: I’m still seeing HTTP 302 for the VRT (even now)

nemo794:
I just now followed the CLI instructions here for setting up AWS environment variables and then subsetting the VRT, and it worked correctly: https://nisar-docs.asf.alaska.edu/nisar-dem/#subsetting-via-vsis3

wildintellect:
Right what's missing is passing the correct GDAL auth headers in the https requests. Maybe @jsignell has an example handy of using EDL+rasterio

wildintellect:
I think it's something like https://gis.stackexchange.com/a/461792
But why are we even messing with the https version when the S3 version works fine and is better for MAAP anyways?

jsignell:
> Right what's missing is passing the correct GDAL auth headers in the https requests. Maybe @jsignell has an example handy of using EDL+rasterio

I don't have something at hand. I would look to the openscape docs to find the right incantation

chuckwondo:
> I think it's something like https://gis.stackexchange.com/a/461792 But why are we even messing with the https version when the S3 version works fine and is better for MAAP anyways?

Good question, and yes, the advice in that post should do the trick.

Further, the reason that the approach that the OP used (i.e., basic auth) won't work is precisely because of the redirection that occurs. Using basic auth in this case is not only insecure (because the creds get sent to the wrong server), but also ineffective because they get stripped on redirection, and thus never reach the intended host.

Including https might be good for completeness, in case somebody attempts to use things outside of an AWS env, but to Alex's point, perhaps we skip this (unless somebody specifically requests it, in which case this can be added later).

Regarding the S3 use case (and actually, for the https case as well, if kept or added later), please use [`rasterio.Env(...)`](https://rasterio.readthedocs.io/en/latest/api/rasterio.env.html#rasterio.env.Env) as a context manager surrounding the call to `open` rather than setting env vars in the code. See the [full list of GDAL config options](https://gdal.org/en/stable/user/configoptions.html#global-configuration-options) that you can pass as kwargs to `rasterio.Env()`.

HarshiniGirish:
@wildintellect  up for review, stuck to just s3 access here.

docs/source/science/NISAR/NISAR_DEM.ipynb
```diff
@@ -0,0 +1,367 @@
+{
```
hrodmn, line 1: <p>As <span class="mention" data-value="chuckwondo" data-denotation-char="@">﻿<span contenteditable="false">@chuckwondo</span>﻿</span> suggested, it is better to use rasterio's AWSSession and the Env context manager to handle AWS credentials. You can delete this chunk and use the suggestion below!</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/562/discussion/#comment-2887296433'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='13'/>

hrodmn, line 1: <p>Please update this to use rasterio's AWSSession:</p><p> </p>

<pre class="ql-syntax" spellcheck="false">def subset_to_geotiff(dataset_path: str, bbox, out_tif: str, creds: dict):
&nbsp; &nbsp; """Read a window from a raster dataset and write it to a local GeoTIFF."""


&nbsp; &nbsp; # Create the AWS Session object
&nbsp; &nbsp; session = AWSSession(
&nbsp; &nbsp; &nbsp; &nbsp; aws_access_key_id=creds["accessKeyId"],
&nbsp; &nbsp; &nbsp; &nbsp; aws_secret_access_key=creds["secretAccessKey"],
&nbsp; &nbsp; &nbsp; &nbsp; aws_session_token=creds["sessionToken"],
&nbsp; &nbsp; &nbsp; &nbsp; region_name="us-west-2",
&nbsp; &nbsp; &nbsp; &nbsp; requester_pays=True,&nbsp; # Handles AWS_REQUEST_PAYER logic
&nbsp; &nbsp; )


&nbsp; &nbsp; with (
&nbsp; &nbsp; &nbsp; &nbsp; rasterio.Env(session=session),
&nbsp; &nbsp; &nbsp; &nbsp; rasterio.open(dataset_path) as src,
&nbsp; &nbsp; ):
&nbsp; &nbsp; &nbsp; &nbsp; minx, miny, maxx, maxy = bbox
&nbsp; &nbsp; &nbsp; &nbsp; win = from_bounds(minx, miny, maxx, maxy, transform=src.transform)


&nbsp; &nbsp; &nbsp; &nbsp; data = src.read(1, window=win)
&nbsp; &nbsp; &nbsp; &nbsp; new_transform = src.window_transform(win)


&nbsp; &nbsp; &nbsp; &nbsp; profile = src.profile.copy()
&nbsp; &nbsp; &nbsp; &nbsp; profile.update(
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; driver="GTiff",
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; height=data.shape[0],
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; width=data.shape[1],
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; transform=new_transform,
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; count=1,
&nbsp; &nbsp; &nbsp; &nbsp; )


&nbsp; &nbsp; with rasterio.open(out_tif, "w", **profile) as dst:
&nbsp; &nbsp; &nbsp; &nbsp; dst.write(data, 1)


</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/MAAP-Project/maap-documentation/pull/562/discussion/#comment-2887296438'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='15'/>

hrodmn (changes requested):
This is cool! I second @chuckwondo's suggestion that we should steer users towards using rasterio's `AWSSession` and `rasterio.Env` to handle S3 credentials when opening files.

Is writing a subset geotiff really what users need? If so this is great, but I would recommend using rioxarray to load a subset into memory rather than writing the subset to disk.

Something like this would be a nice substitute for the write-to-disk approach:

```python
import rioxarray
import rasterio
from rasterio.session import AWSSession

session = AWSSession(
    aws_access_key_id=creds["accessKeyId"],
    aws_secret_access_key=creds["secretAccessKey"],
    aws_session_token=creds["sessionToken"],
    region_name="us-west-2",
    requester_pays=True
)


with rasterio.Env(session=session):
    rds = rioxarray.open_rasterio(S3_VRT_S3URI, chunks="auto", mask_and_scale=True)
    clipped = rds.rio.clip_box(*BBOX)

# plot the array
clipped.plot.imshow()
```

wildintellect:
@hrodmn I agree, but also expect users will write out subsets too - they like to build on disk datacubes/stacks for later. This example doesn't need to show that though, and note, we should always show writing to a cloud optimized format. So +1 drop the writing out from this example, but maybe link to another one where writing from a Xarray to COG is done.

hrodmn:
> maybe link to another one where writing from a Xarray to COG is done.

That's a one-liner with rioxarray!
```python
clipped.rio.to_raster("/path/to/output.tif", driver="COG")
```

wildintellect:
Can you verify that maintains the projection information correctly? I recall needing to manually set crs in the past.

HarshiniGirish:
yes @wildintellect , projection information is being maintained correctly.

HarshiniGirish:
Thankyou @hrodmn. I've made the changes. up for review again!

hrodmn (approved):
Looks great, thanks @HarshiniGirish !

PR #567, opened 2026-03-05
Pointing to VEDA docs for SSH instructions 

grallewellyn:


wildintellect (approved):
Good enough for now. We'll probably want a little more explanation later.

PR #577, opened 2026-05-04
Merge Hub Documentation

smk0033:
With more users on the Hub and the upcoming phasing out of the ADE, Hub docs should be more visible to the general user. This PR is to merge the new Hub documentation to develop, and then we can do a release to the master branch.


wildintellect:
The key thing actually is to drive new users to sign up via the Hub process as it collects better information. But it also positions it as the default, as we work our way through migration.

grallewellyn (approved):

wildintellect:
@omshinde @HarshiniGirish can you resolve the ESA_CCI conflict?

HarshiniGirish:
@wildintellect  the conflict is resolved.

PR #583, opened 2026-05-11
ADE Docs Notice

smk0033:
Added a notice to the top of the docs home page to inform users where the last release is of ADE-only documentation is.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -5,6 +5,8 @@
 
 Welcome to the MAAP User Documentation!
 =======================================
+.. attention::
+    MAAP is transitioning from the ADE to Jupyterhub, and new Jupyterhub documentation is being added. For access to *ADE only* documentation, please see the `last-ade-release <https://docs.maap-project.org/en/last-ade-release/>`_. 
 
 .. toctree::
   :maxdepth: 1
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -5,6 +5,8 @@
 
 Welcome to the MAAP User Documentation!
 =======================================
+.. attention::
+    MAAP is transitioning from the ADE to Jupyterhub, and new Jupyterhub documentation is being added. For access to *ADE only* documentation, please see the `last-ade-release <https://docs.maap-project.org/en/last-ade-release/>`_.
```
wildintellect: It's a little confusing because both use Jupyterhub. We refer to them as MAAP ADE and MAAP Hub.

So maybe

> MAAP is transitioning from the MAAP ADE to MAAP Hub for interactive JupyterHub coding environments. The documentation not reflects the MAAP Hub as the default, if you need to look up information specifice to MAAP ADE please see ...
> 
> Documentation around all other aspect of MAAP remains unchanged, including use of maap-py and DPS.

wildintellect (approved):
Great, we'll just make a note to ourselves that we can remove this notice once we turn off the ADE.

PR #592, opened 2026-07-30
Add documentation for the universal sign up page and PATs

bsatoriu:
- Introduced a new section detailing Personal Access Tokens for authenticating with the MAAP API.
- Explained the creation process, usage with maap-py, and management of tokens.
- Included notes on NASA and ESA token distinctions and security best practices.


grallewellyn (approved):
Looks good to me

wildintellect:
I think the only question that might come up, is when people don't make an EDL login before their first attempt, will need to come back to the signup page afterwards manually.

bsatoriu:
> when people don't make an EDL login before their first attempt

@wildintellect I'm not sure what you mean here. What are the sequence of events in this scenario?

wildintellect:
I went to the signup page, it sent me to EDL. Since I didn't have an account I had to click Register, after registration it dead ends at account verification for EDL there is no redirect back to MAAP registration. So I had to start over, and on the second time because I have an EDL account, now I can actually sign up for MAAP.

bsatoriu:
Got it. Thanks for the breakdown. I'll update the docs.

PR #593, opened 2026-07-31
Refactor Earthdata Login account signup section for clarity

bsatoriu:


wildintellect:
Do we need to put anything about ESA users in there, or wait until the ESA id works for EDL?

wildintellect (approved):

bsatoriu:
> Do we need to put anything about ESA users in there, or wait until the ESA id works for EDL?

Good question--I think waiting until EDL federation is available makes sense. As soon as it's ready, I'll create a new PR.

PR #598, opened 2026-08-27
docs: make DPS STAC guide visible

### LINKED ISSUE #597: Make DPS STAC guide visible
We have a guide for using the DPS STAC auto-cataloging feature but it is not listed in the navigation page. We need to add it to the technical tutorials!

Guide link: https://github.com/MAAP-Project/maap-documentation/blob/develop/docs/source/technical_tutorials/dps_tutorial/dps_stac_metadata.ipynb

hrodmn:
Add the DPS STAC metadata guide to the Technical Tutorials navigation.

resolves #597

### DIFF
--- a/docs/source/technical_tutorials.rst
+++ b/docs/source/technical_tutorials.rst
@@ -10,6 +10,7 @@ This section of the documentation includes hands-on tutorials with example algor
 
 
   technical_tutorials/dps_tutorial/dps_tutorial_demo.ipynb
+  technical_tutorials/dps_tutorial/dps_stac_metadata.ipynb
   technical_tutorials/searching.rst
   technical_tutorials/visualizing.rst
   technical_tutorials/accessing.rst


wildintellect:
Should we update some of the text before we list it? Maybe that's a minor for later thing.
- Link in description are broken (translating to relative urls)
- Why to catalog description is a bit inverted.
- needs to be more clear that more than COGs can be cataloged

Side Ticket:
- Make sure STAC browser works with DPS catalog and include how to browse items once added

cc: @sujen1412

wildintellect:
I mentioned to @sujen1412 that the only blocker I have on this, now is that we should tag the DPS STAC as Beta or Experimental with a statement that migration of records to future changes will be possible but might not be automatic. However the effort to use is low and the benefits can be great.

@sujen1412 also asked a question, what about cataloging existing items (e.g. files imported to MAAP), and I mentioned the example algorithm is actually a catalog only operation on existing files.

PR #600, opened 2026-09-11
first pass of migrating to MyST / Jupyterbook, moving to MyST structure

batpad:
Refs https://github.com/NASA-IMPACT/active-maap-sprint/issues/1461

Opening this as a draft - did a first pass with Claude to create deterministic migration scripts and move to the Jupyterbook structure. Went better than expected and testing locally, it seemed like things all(?) seem to work - the R examples, Leaflet maps, etc. all seemed to render fine.

There's a bunch of agent generated docs about the migration including compromises made etc. in migration/agent-docs/

Trying to continue to use the ReadTheDocs structure to not have to change hosting in the same pass, but let's see how that works. If the RTD previews don't work, I'll figure just pushing the static HTML folder built by MyST somewhere so we can see what this looks like and start identifying any inconsistencies, etc. @choldgraf may ask for your help with some Netlify slinging to get a quick previewable URL up.

This first pass looks promising to me in that we may be able to move to Jupyterbook without changing a lot of infrastructure or making any drastic changes to the visual look and feel, so it maybe plausible to "just move" fairly soon to reproduce the same docs while switching out the underlying build technology, and then iteratively work on improvements to fully utilize newer MyST Jupyterbook features / migrate to nicer widgets, etc.

Will write more detailed notes about the PR and some choices made - agent generated AI-slop docs live in migration/agent-docs/ and should cover most things until I'm able to gather thoughts and write a more thoughtful note.

cc @zacdezgeo @choldgraf @wildintellect


batpad:
While we figure out getting this to a preview URL, here's a screenshot from local showing the docs re-built using Jupyterbook, showing an R example with a map working:

image

choldgraf:
I would expect readthedocs to work just fine - we use it for jupyter book's docs and it has been fine. So lemme know if you need me to help spot check anything

wildintellect:
ah interesting, minor config change https://about.readthedocs.com/tools/jupyter-book/

Just not sure if we can do that and not break the existing site. FYI, we specifically configure which branches are built on the readthedocs hosting.

wildintellect:
It's working, I just had to activate the build https://docs.maap-project.org/en/feat-migrate-to-jupyterbook/

zacdezgeo:
@batpad ; ah I love the look and feel of the new docs 🥳  the only thing I caught which doesn't perfectly match which the docs is sub section headers/sub lists. for example, in the left panel index, we have sections of the tutorials and examples listed in the old docs:
image

Similarly, when we index pages at the start, we aren't listing sub-sections like in the older docs: 
_[old docs]_
image

_[new docs]_
image

Personally, I prefer the new version of the docs for the index in the left panel (not listing section headers). However, I could see us wanting to have finer grained sections at the top of pages like the getting started one. @wildintellect, any thoughts here?

batpad:
Thanks @zacdezgeo ! Let's maybe wait for @wildintellect to comment on what the ideal menu sections / subsections layout should be, and then hopefully @choldgraf is able to help us with how to achieve that with the Jupyterbook theme.

@zacdezgeo if there's concrete decisions on how we would like subsections represented and changes to make in the migration branch, could you make focussed tickets that describe the current behaviour and desired behaviour a bit exactly?

agoose77:
@batpad it's not currently possible to show page-level headings in the primary sidebar (left menu). 

We can set the table of contents depth for the in-page ToC, though: https://mystmd.org/guide/directives#directive-toc-depth

wildintellect:
I'm ok with the sub headings in the break out on the right. The key is that sub page navigation exists, not that it must be on the primary nav on the left (the only place the current system allowed).

choldgraf:
Yeah - that was an intentional design decision for Jupyter Book and MyST, since it seems to follow the more common convention in web tech these days (to put the in-page content headers in a dropdown to the right, and the "between pages" navigation in a sidebar to the left). That's how the pydata-sphinx-theme and the sphinx-book-theme operated as well. Users these days seem to "expect" the in-page headers to be clickable on the right 🤷

batpad:
> We can set the table of contents depth for the in-page ToC, though: https://mystmd.org/guide/directives#directive-toc-depth

Hmm @agoose77 that did not seem to work for me to achieve the effect on https://docs.maap-project.org/en/latest/getting_started.html - setting `depth` seems to affect "levels of pages" and not "headings inside those pages" if that makes sense? It's unclear to me how to get those in-page sub-headings for the Getting Started page for example to appear as the old site without a custom plugin.

batpad:
The version switcher now appears: https://docs.maap-project.org/en/feat-migrate-to-jupyterbook/ - it was a MyST / React re-render thing that was causing it to hide.

The problem is that the URL structure changed with Jupyterbook vs Sphinx, so if navigating using the version switcher from any page other than the home-page, one gets a 404. So don't think this quite works. I think there is an option of setting up redirects if we were using a paid version of Read the Docs.

So I think options probably are:

 - Explore using https://github.com/DiamondLightSource/myst-version-switcher-plugin and not the RTD version switcher, but am not sure how that would work with different versions being built using different mechanisms
  - Get a paid RTD account - am pretty sure this can be fixed with setting up redirects with the paid plan
   - Move off RTD entirely onto our own hosting setup with s3 or so - does not necessarily make things easier, just lets us architect and build something for our specific needs without facing any artificial constraints imposed by the platform
   - Figure out some hack to get the current version switcher to at-least "kindof" work - not sure what options might be
   
  Am really not sure if I have a preference here. @wildintellect @zacdezgeo I think the version switcher is likely the biggest blocker to being able to move? Let's maybe chat about this? I can play around a bit more and see if we have any other options.

(+cc @sunu )

batpad:
One "hack" for the version switcher that would work:

We create a custom 404 page, so if someone does a version switch on a URL that does not exist for the past versions, we can check which version they were going to on the custom 404 page, and link (or redirect) to the home-page of that version. So we'd lose the functionality of being able to jump to another version from an arbitrary page, but things would not be broken and you can still use the version switcher to get to previous versions of the docs.

I can try implementing this so we can see how it works - I feel like this should be good enough to unblock, and then we can discuss if we want to move off RTD anyways, etc.

